PR #10, opened 2021-02-02
Removes standalone agent functionality, and prepares the new repo for its first release.

djaglowski:
- Module name now matches repo name exactly, so it can be imported elsewhere
- Prepare changelog for an initial release
- Removes standalone agent functionality
- Simplifies modules
  - Combines `parser/syslog` and `input/windows` modules into the primary module
  - Removes output operators that were previously separate modules
  - Leaves `input/k8sevent` and `transformer/k8smetadata` as separate modules for now. These two have extensive dependencies and their usefulness in the collector needs to be discussed before merging or removing.


operator/builtin/parser/syslog/go.mod
```diff
@@ -1,11 +0,0 @@
-module github.com/opentelemetry/opentelemetry-log-collection/operator/builtin/parser/syslog
```
tigrannajaryan, line 1: Are you eliminating the extra modules and making the library all one module?

djaglowski: Only for syslog, since it's on the near-term roadmap.

We can leave it if you prefer. It's just more process to manage it separately so I thought it would make sense to bring it into the primary module.

Sorry - should have called attention to this in the pull request notes either way.

djaglowski: I suppose Windows Event Log should be treated the same as syslog, whichever way is decided. Do you have a preference?

tigrannajaryan: No preference, works either way, just make it consistent.

tigrannajaryan (approved):

PR #25, opened 2021-02-25
file_input add mapstructure support

### LINKED ISSUE #14: Add mapstructure support to file_input operator configuration
This will allow explicit embedding in the collector's `filelogreceiver`.
- This change should perhaps be preceded by enhanced test cases on the `file_input` configuration

WPH95:
- [x] test
- mapstructure
	- [x] FileInputConfig
	- [x] InputConfig	
	- [x] WriterConfig
	- [x] BaseConfig
	- [x] WriteTo

ref https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2334
resolved #14


operator/builtin/input/file/config_test.go
```diff
@@ -0,0 +1,91 @@
...
+	dc := &mapstructure.DecoderConfig{Result: &actual, DecodeHook: mapstructure.TextUnmarshallerHookFunc()}
+	ms, err := mapstructure.NewDecoder(dc)
+	require.NoError(t, err)
+	err = ms.Decode(input)
+	require.NoError(t, err)
+	require.Equal(t, except, &actual)
+}
+
+func TestMapStructureDecodeConfig(t *testing.T) {
+	except := NewTestInputConfig()
```
djaglowski, line 59: nit: This should be `expect`

djaglowski (commented):
This looks good to me. However, I would like to be very cautious with this change.

Before we can be confident that all unmarshaling is the same between previous `yaml` and now `mapstructure`, I believe we need to include some additional tests.

1. We should have some "golden" config files, which are unmarshaled using both `yaml` and `mapstructure`, to ensure that the resulting config is exactly the same.
2. Especially important, we need to ensure that custom types (those defined in `helper` package) are unmarshaled the same in both cases.

I have added another ticket here, which I will take myself: #33

operator/helper/bytesize.go
```diff
@@ -34,6 +34,14 @@ func (h *ByteSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return h.unmarshalShared(unmarshal)
 }
 
+func (h *ByteSize) UnmarshalText(text []byte) (err error) {
```
djaglowski, line 37: I believe we should expand this test loop to also unmarshal the raw values using mapstructure, and validate the results are the same. 

Very rough implementation:

```
for i, tc := range cases {
		t.Run(strconv.Itoa(i), func(t *testing.T) {
			var bsViaYaml ByteSize
			err := yaml.Unmarshal([]byte(tc.input), &bsViaYaml)
			if tc.expectError {
				require.Error(t, err)
				return
			}

			require.NoError(t, err)
			require.Equal(t, tc.expected, bs)

                        // And also same result if using mapstructure
                        var bsViaMapstructure ByteSize
		        err := mapstructure.Decode([]byte(tc.input), &bsViaMapstructure)
                        if tc.expectError {
				require.Error(t, err)
				return
			}

			require.NoError(t, err)
			require.Equal(t, tc.expected, bs)
		})
	}
```

operator/helper/duration.go
```diff
@@ -72,6 +72,10 @@ func (d *Duration) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return err
 }
 
+func (d *Duration) UnmarshalText(text []byte) error {
```
djaglowski, line 75: Similar to comment on `bytesize_test.go`, we should enhance tests to ensure unmarshaling is the same between yaml and mapstructure in `duration_test.go`.

djaglowski:
> We should have some "golden" config files, which are unmarshaled using both `yaml` and `mapstructure`, to ensure that the resulting config is exactly the same.

@WPH95 I have merged #40, which I believe establishes a reasonable pattern for ensuring consistency between `yaml` and `mapstructure` unmarshaling. Would you mind pulling this in and including mapstructure validation of the same cases?

WPH95:
> > We should have some "golden" config files, which are unmarshaled using both `yaml` and `mapstructure`, to ensure that the resulting config is exactly the same.
> 
> @WPH95 I have merged #40, which I believe establishes a reasonable pattern for ensuring consistency between `yaml` and `mapstructure` unmarshaling. Would you mind pulling this in and including mapstructure validation of the same cases?

np, will try it tomorrow, #40 look great

operator/builtin/input/file/config_test.go
```diff
@@ -230,8 +232,115 @@ func configFromFileViaYaml(t *testing.T, file string) (*InputConfig, error) {
...
-// TODO func configFromFileViaMapstructure(t *testing.T, file string) (*InputConfig, error) {}
+func configFromFileViaMapstructure(file string) (*InputConfig, error) {
+	bytes, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, fmt.Errorf("could not find config file: %s", err)
+	}
+
+	raw := map[string]interface{}{}
+
+	if err := yaml.Unmarshal(bytes, raw); err != nil {
```
djaglowski, line 243: 👍

djaglowski (approved):
This looks excellent.

operator/helper/mapstructure.go
```diff
@@ -0,0 +1,38 @@
+package helper
```
djaglowski, line 1: Just need a license

PR #30, opened 2021-02-26
base ParserConfig add mapstructure support

### LINKED ISSUE #18: Add mapstructure support to parser configurations
Similar to #14, #15, #16, #17, but will have a different effect on the collector receivers.

Currently, the collector unmarshals these configs into an `[]map[string]interface{}`, and then uses yaml marshaling. Since we don't know which operators will be included and in what order, this might not have as much of a benefit as the other tickets, but there is likely a better design than the current.

General guidelines:
- This change should perhaps be preceded by enhanced test cases on the parser configurations

WPH95:
depend #25 
resolved #18


WPH95:
#58 is merged close this

PR #35, opened 2021-03-01
Operator tcp input scanner buffer

jsirianni:
Added optional max_buffer_size parameter to tcp_input operator. This is useful for situations where TCP input is handling payloads larger than the default scanner's buffer. 

For more context, see:
Original Stanza issue https://github.com/observIQ/stanza/issues/255
Original Stanza PR https://github.com/observIQ/stanza/pull/256


djaglowski (approved):
LGTM. Thanks @jsirianni

PR #40, opened 2021-03-01
Golden config pattern

### LINKED ISSUE #33: Unmarshaling tests with golden files
Before switching to `mapstructure`, we should harden our existing config expectations by establishing a golden file pattern. This will allow for higher confidence in testing of unmarshaling functionality because the golden files can be unmarshaled using both `yaml` and `mapstructure` and then compared to the same expected result.

djaglowski:
Establishes a pattern that resolves #33.

This can be pulled into #25 and extended in order to unblock.


jsirianni (approved):
Looks good, this is awesome 👍

PR #57, opened 2021-03-09
Add mapstructure support and expand test cases for "Duration" type

Mrod1598:
Expanded mapstructure support to Duration. This involved modifying how strings are handled within duration to adequately cover string values, specifically those without a unit attached. Also, expanded test cases to cover more scenarios.


operator/helper/duration_test.go
```diff
@@ -16,48 +16,153 @@ package helper
...
-		input    string
-		expected Duration
+		name        string
+		input       string
+		expected    Duration
+		expectError bool
 	}{
+		{
+			"simple nanosecond",
+			`"10ns"`,
```
djaglowski, line 37: We should also have tests with unquoted values

operator/helper/duration_test.go
```diff
@@ -104,3 +236,26 @@ func TestParseDurationRoundtrip(t *testing.T) {
 		})
 	}
 }
+
+func ToTimeHookFunc() mapstructure.DecodeHookFunc {
```
djaglowski, line 240: @Mrod1598 I think this must have been left over from an earlier implementation

djaglowski (approved):
LGTM

PR #58, opened 2021-03-11
all builtin parser add mapstructure support

### LINKED ISSUE #18: Add mapstructure support to parser configurations

WPH95:
wait for #30 
resolved #18 

---

### test 
- [x] json 
- [x] regex
- [x] serverity
- [x] syslog
- [x] time
- [x] url


operator/helper/mapstructure.go
```diff
@@ -50,3 +50,12 @@ func JSONUnmarshalerHook() mapstructure.DecodeHookFunc {
 		return to.Interface(), nil
 	}
 }
+
+func UnmarshalMapstructure(input interface{}, result interface{}) error {
```
djaglowski, line 54: 👍

djaglowski (approved):
Looks great!

PR #60, opened 2021-03-12
use opentelemetry collect standard tls config instead of tc tls config

### LINKED ISSUE #59: Improve TLS config, use standard TLS config
use standard TLS config 
https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/configtls

to instead of 
https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/builtin/input/tcp/tcp.go#L63

WPH95:
resolved #59


operator/helper/tls.go
```diff
@@ -0,0 +1,39 @@
...
+
+package helper
+
+import (
+	"github.com/mitchellh/mapstructure"
+	"go.opentelemetry.io/collector/config/configtls"
+)
+
+type TLSServerConfig struct {
+	*configtls.TLSServerSetting
```
djaglowski, line 23: Should this have `mapstructure` / `json` / `yaml` tags?

djaglowski (approved):
Looks good to me. One minor comment.

operator/helper/tls.go
WPH95: yeah, it's should be, let me add it

PR #64, opened 2021-03-16
Json golden Config unmarshalling

Mrod1598:
Add golden configs to further test unmarshalling for json_parser


operator/builtin/parser/json/testdata/on_error_drop.yaml
```diff
@@ -0,0 +1,2 @@
+type: json_parser
+on_error: "drop"
```
djaglowski, line 2: The quotes are necessary on string values in yaml, but they don't do any harm either. Typically, we should not include the quotes, unless we are testing something specifically relating to them.

djaglowski (approved):
LGTM

operator/builtin/parser/json/testdata/on_error_drop.yaml
djaglowski: @Mrod1598 Can you also do this for the string values in other test cases? (Unless there is a reason they need the quotes)

djaglowski (approved):
Nice work!

PR #65, opened 2021-03-16
Add regex parser golden configs

Mrod1598:
added more testing for unmarshalling yaml files for regex_parser operator.


djaglowski (approved):
LGTM

PR #66, opened 2021-03-16
Add golden config testing to metadata operator

Mrod1598:
Added golden config testing that touches all of the fields.


djaglowski (approved):
LGTM

PR #76, opened 2021-03-27
Add trace attributes and parser (trace_id, span_id and trace_flags)

alexvanboxel:
This PR is to ask for feedback from the community. Locally I have a hacked OTEL collector running that uses the new trace attribute, transforms it to OTLP Log, and then pushes it to Google Logging.

I think this is the right place to introduce it as trace is a core component of OTLP.

If this is ok, I'll create tests and make it a proper commit.

@djaglowski

### DIFF
--- a/entry/entry.go
+++ b/entry/entry.go
@@ -48,6 +48,9 @@ type Entry struct {
 	Attributes   map[string]string `json:"attributes,omitempty"    yaml:"attributes,omitempty"`
 	Resource     map[string]string `json:"resource,omitempty"      yaml:"resource,omitempty"`
 	Body         interface{}       `json:"body"                    yaml:"body"`
+	TraceId      []byte            `json:"trace_id,omitempty"      yaml:"trace_id,omitempty"`
+	SpanId       []byte            `json:"span_id,omitempty"       yaml:"span_id,omitempty"`
+	TraceFlags   []byte            `json:"trace_flags,omitempty"   yaml:"trace_flags,omitempty"`
 }
 
 // New will create a new log entry with current timestamp and an empty body.
@@ -196,5 +199,8 @@ func (entry *Entry) Copy() *Entry {
 		Attributes:   copyStringMap(entry.Attributes),
 		Resource:     copyStringMap(entry.Resource),
 		Body:         copyValue(entry.Body),
+		TraceId:      copyByteArray(entry.TraceId),
+		SpanId:       copyByteArray(entry.SpanId),
+		TraceFlags:   copyByteArray(entry.TraceFlags),
 	}
 }
--- a/operator/builtin/parser/trace/trace_test.go
+++ b/operator/builtin/parser/trace/trace_test.go
@@ -1,0 +1,158 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package trace
+
+import (
+	"encoding/hex"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/open-telemetry/opentelemetry-log-collection/entry"
+	"github.com/open-telemetry/opentelemetry-log-collection/testutil"
+)
+
+func TestDefaultParser(t *testing.T) {
+	traceParserConfig := NewTraceParserConfig("")
+	_, err := traceParserConfig.Build(testutil.NewBuildContext(t))
+	require.NoError(t, err)
+}
+
+func TestTraceParserParse(t *testing.T) {
+	cases := []struct {
+		name           string
+		inputRecord    map[string]interface{}
+		expectedRecord map[string]interface{}
+		expectErr      bool
+		traceId        string
+		spanId         string
+		traceFlags     string
+	}{
+		{
+			"AllFields",
+			map[string]interface{}{
+				"trace_id":    "480140f3d770a5ae32f0a22b6a812cff",
+				"span_id":     "92c3792d54ba94f3",
+				"trace_flags": "01",
+			},
+			map[string]interface{}{},
+			false,
+			"480140f3d770a5ae32f0a22b6a812cff",
+			"92c3792d54ba94f3",
+			"01",
+		},
+		{
+			"WrongFields",
+			map[string]interface{}{
+				"traceId":    "480140f3d770a5ae32f0a22b6a812cff",
+				"traceFlags": "01",
+				"spanId":     "92c3792d54ba94f3",
+			},
+			map[string]interface{}{
+				"traceId":    "480140f3d770a5ae32f0a22b6a812cff",
+				"spanId":     "92c3792d54ba94f3",
+				"traceFlags": "01",
+			},
+			false,
+			"",
+			"",
+			"",
+		},
+		{
+			"OnlyTraceId",
+			map[string]interface{}{
+				"trace_id": "480140f3d770a5ae32f0a22b6a812cff",
+			},
+			map[string]interface{}{},
+			false,
+			"480140f3d770a5ae32f0a22b6a812cff",
+			"",
+			"",
+		},
+		{
+			"WrongTraceIdFormat",
+			map[string]interface{}{
+				"trace_id":    "foo_bar",
+				"span_id":     "92c3792d54ba94f3",
+				"trace_flags": "01",
+			},
+			map[string]interface{}{},
+			true,
+			"",
+			"92c3792d54ba94f3",
+			"01",
+		},
+		{
+			"WrongTraceFlagFormat",
+			map[string]interface{}{
+				"trace_id":    "480140f3d770a5ae32f0a22b6a812cff",
+				"span_id":     "92c3792d54ba94f3",
+				"trace_flags": "foo_bar",
+			},
+			map[string]interface{}{},
+			true,
+			"480140f3d770a5ae32f0a22b6a812cff",
+			"92c3792d54ba94f3",
+			"",
+		},
+		{
+			"AllFields",
+			map[string]interface{}{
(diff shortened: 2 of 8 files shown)


alexvanboxel:
FYI: The end-to-end config I have success with (where this is part of):

config for Java (logback): 

```
    <appender name="JSON" class="ch.qos.logback.core.rolling.RollingFileAppender">
        <file>${LOG_JSON_PATH}/xxx.jsonl</file>
        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
            <fileNamePattern>${LOG_JSON_PATH}/dgc${LOG_FILE_NAME_PATTERN}.jsonl</fileNamePattern>
            <maxFileSize>10MB</maxFileSize>
            <maxHistory>5</maxHistory>
            <totalSizeCap>50MB</totalSizeCap>
        </rollingPolicy>

        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
            <lineSeparator>UNIX</lineSeparator>
            <timestampPattern>[UNIX_TIMESTAMP_AS_NUMBER]</timestampPattern>
            <fieldNames>
                <timestamp>timestamp_ms</timestamp>
                <version>[ignore]</version>
                <levelValue>[ignore]</levelValue>
            </fieldNames>
        </encoder>
    </appender>

```

otel-collector

```
  filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: "json_parser"
        timestamp:
          parse_from: "timestamp_ms"
          layout_type: "epoch"
          layout: "ms"
        severity:
          parse_from: "level"
        trace:
          parse_trace_from: "trace_id"
          parse_span_from: "span_id"
          parse_flags_from: "trace_flags"
```

entry/entry.go
```diff
@@ -196,5 +199,7 @@ func (entry *Entry) Copy() *Entry {
 		Attributes:   copyStringMap(entry.Attributes),
 		Resource:     copyStringMap(entry.Resource),
 		Record:       copyValue(entry.Record),
+		TraceId:      entry.TraceId,
+		SpanId:       entry.SpanId,
```
djaglowski, line 203: Should `TraceFlags` also be copied here?

operator/helper/trace.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func parse(entry *entry.Entry,field *entry.Field, to *entry.Field) []byte {
+	value, ok := entry.Delete(field)
+	if !ok {
+		return nil
+	}
+
+	data, err := hex.DecodeString(fmt.Sprintf("%v",value))
```
djaglowski, line 58: I don't have a lot of experience working with traces, so let me ask a very basic question:

Is it a safe assumption that `TraceID`, `SpanID`, and `TraceFlags` should always be hex encoded? In other words, is this implementation specific to any particular tracing technology or standard?

djaglowski (commented):
Looks great overall - just a couple questions

operator/helper/trace.go
tigrannajaryan: Hex-encoding requirement and the specific names of the fields is part of Otel spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#trace-context-in-legacy-formats

djaglowski: So this implementation will allow us to parse tracing data from logs that adhere to the OTel spec. 

Certainly this is an improvement upon what is supported today, but isn't the purpose of the library to ingest logs from all sources? 

I'm just trying to understand whether we are solving this in a generalized way, or if we are making an assumption about log format. Sounds like it's the latter, in which case I think we should consider this implementation preliminary, and plan to generalize it later. 

Maybe something like:

```
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: "json_parser"
        timestamp:
          parse_from: "timestamp_ms"
          layout_type: "epoch"
          layout: "ms"
        severity:
          parse_from: "level"
        trace_id:
          parse_from: "trace_id"
          encoding: "hex" // default is "hex", for now hardcoded as "hex"
        span_id:
          parse_from: "span_id"
          encoding: "hex" // default is "hex", for now hardcoded as "hex"
        trace_flags:
          parse_from: "trace_flags"
          encoding: "hex" // default is "hex", for now hardcoded as "hex"
```

djaglowski: I would even go so far as to say that the config in this PR needs to be changed in order to accomodate the above. We don't necessarily need to support the `encoding` parameter yet, but we should anticipate that it will be added later:

```
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: "json_parser"
        timestamp:
          parse_from: "timestamp_ms"
          layout_type: "epoch"
          layout: "ms"
        severity:
          parse_from: "level"
        trace_id:
          parse_from: "trace_id"
        span_id:
          parse_from: "span_id"
        trace_flags:
          parse_from: "trace_flags"
```

tigrannajaryan: @djaglowski I agree. Let's just make the defaults compliant with Otel spec.

alexvanboxel: Mmm, in the suggested config the trace/span/flags are split, and if I understand it correctly that would mean 3 different operators? That also means that if you want to default you need to specify:

3 operators for defaults:

```
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: "json_parser"
        trace_id: {}
        span_id: {}
        trace_flags: {}
```

1 operator for defaults:

```
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: "json_parser"
        trace: {}
```

Anyone strong opinions?

djaglowski: I see your point about what I suggested. I think it can be done with 1 operator though, by pushing it down one level, as shown in your last suggestion:
```yaml
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: json_parser
        trace: {}
```

Which, accounting for default values would give an effective configuration:
```yaml
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: json_parser
        trace:
          trace_id:
            parse_from: trace_id
          span_id:
            parse_from: span_id
          trace_flags:
            parse_from: trace_flags
```

One could partially customize this, as in:
```yaml
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: json_parser
        trace:
          span_id:
            parse_from: my_custom_span_id
```

And this would also allow us to extend as necessary in the future:
```yaml
filelog:
    include: [ /.../logs/*.jsonl ]
    operators:
      - type: json_parser
        trace:
          trace_id:
            parse_from: trace_id
            encoding: hex
          span_id:
            parse_from: my_custom_span_id
            encoding: md5
          trace_flags:
            parse_from: my_trace_flags_array
            encoding: array
            some_other_config_thing: my_value
```

The above may need some tweaking to make sense for traces, but my primary concern is ensuring that we can extend the design later to support trace parsing for non-OTel compliant logs.

alexvanboxel: Alright, that makes sense. I'll do it like that (skipping the customization for now, defaulting to the otel spec)

alexvanboxel:
Rebased against main, squashed into a single commit. I think I took all the concerns into accounts and added tests.

operator/builtin/parser/trace/trace_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestTraceParserParse(t *testing.T) {
+	cases := []struct {
+		name           string
+		inputRecord    map[string]interface{}
+		expectedRecord map[string]interface{}
+		expectErr      bool
+		traceId        string
+		spanId         string
+		traceFlags     string
```
djaglowski, line 41: Are we overspecifying here? It seems like `inputRecord` and `expectedRecord` would be enough to test that we're getting the expected behavior.

operator/builtin/parser/trace/trace_test.go
```diff
@@ -0,0 +1,155 @@
...
+			if len(tc.spanId) == 0 {
+				spanId = nil
+			}
+			traceFlags, _ := hex.DecodeString(tc.traceFlags)
+			if len(tc.traceFlags) == 0 {
+				traceFlags = nil
+			}
+			require.Equal(t, traceId, e.TraceId)
+			require.Equal(t, spanId, e.SpanId)
+			require.Equal(t, traceFlags, e.TraceFlags)
```
djaglowski, line 152: I don't understand why this is necessary. Is it really?

operator/helper/parser.go
```diff
@@ -154,6 +163,10 @@ func (p *ParserOperator) ParseWith(ctx context.Context, entry *entry.Entry, pars
 		severityParseErr = p.SeverityParser.Parse(entry)
 	}
 
+	if p.TraceParser != nil {
+		_ = p.TraceParser.Parse(entry)
```
djaglowski, line 167: We need to handle the error like `time` and `severity`.

djaglowski: Admittedly, the error handling below probably needs to be reworked a bit, but I think it's ok for this PR to just add another individual check:

```
if timeParseErr != nil {
    return p.HandleEntryError(ctx, entry, errors.Wrap(timeParseErr, "time parser"))
}
if severityParseErr != nil {
    return p.HandleEntryError(ctx, entry, errors.Wrap(severityParseErr, "severity parser"))
}
if traceParseErr != nil {
    return p.HandleEntryError(ctx, entry, errors.Wrap(traceParseErr, "trace parser"))
}
```

entry/entry.go
```diff
@@ -196,5 +199,8 @@ func (entry *Entry) Copy() *Entry {
...
+		TraceFlags:   entry.TraceFlags,
```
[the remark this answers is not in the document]
alexvanboxel: Deep copy added, was not sure for nil []byte, added Nil test case and aligned behavior with the rest of the fields

operator/builtin/parser/trace/trace_test.go
alexvanboxel: Reworked a bit to make it more clear. The DecodeString is to make it easier to build the tests

operator/builtin/parser/trace/trace_test.go
alexvanboxel: Euh... well you want to see that the decoding move the traces to the correct fields and parse them into bytes.

djaglowski (commented):
This looks good. Thanks for the idea and implementation.

You've got a merge conflict with `main` due to #88, but that should be pretty straightforward to resolve.

alexvanboxel:
Rebase and Squashed. Once it's merged and the log-collector is released, I'll need to wire up the extra fields in the otel-collector.

djaglowski (approved):
Thanks @alexvanboxel!

Issue #68, opened 2021-03-19
Decompose "restructure" operator


djaglowski:
This operator is essentially a set of related operators bundled together. Because of this, the operator is internally more complicated than necessary, and has proven unnecessarily difficult to maintain / update. It also provides a less consistent user experience than necessary.

Because of this, I suggest that each of the supported "ops" should be made available as a standalone operator. This operator can be left in place or deprecated once equivalent functionality has been proven out.

djaglowski:
@Mrod1598 is planning to undertake this effort.

djaglowski:
Closed via #75, #77, #78, #79, #80, #81

PR #77, opened 2021-03-29
Add Flatten operator

Mrod1598:


operator/builtin/transformer/flatten/flaltten.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Build will build a Flatten operator from the supplied configuration
+func (c FlattenOperatorConfig) Build(context operator.BuildContext) ([]operator.Operator, error) {
+	transformerOperator, err := c.TransformerConfig.Build(context)
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.Contains(c.Field.String(), "$attributes") || strings.Contains(c.Field.String(), "$resource") {
```
Mrod1598, line 52: So, I wanted to note why this works the way it does. When it is unmarshalled it is allows the prefix for "$attributes" and "$resource" through and just saves it as a key. so since it doesn't it catch it during unmarshalling I'm catching it here, Unless we want to change how the bodyfield unmarshalls

docs/operators/flatten.md
```diff
@@ -0,0 +1,99 @@
...
+<br>
+<br>
+
+```yaml
+- type: flatten
+    field: key1
+```
+
+<table>
+<tr><td> Input Body </td> <td> Output Body </td></tr>
```
djaglowski, line 28: Even though this operator only works on the body, let's use the same example format as the others (showing the entire entry).

operator/builtin/transformer/flatten/flaltten.go
djaglowski: This is good

docs/operators/flatten.md
```diff
@@ -0,0 +1,99 @@
+## `flatten` operator
+
+The `flatten` operator flattens a field by moving its children up to the same level as the field.
```
djaglowski, line 3: My intuition was that this would happen recursively. However, this would introduce a lot complications, so I actually like this version, for now at least.

That said, I think we need to call this out explicitly here, since others may also assume that it is a recursive operation.

operator/builtin/transformer/flatten/flatten_test.go
```diff
@@ -0,0 +1,159 @@
...
+				e.Body = map[string]interface{}{
+					"key": "val",
+					"nested": map[string]interface{}{
+						"secondlevel": map[string]interface{}{
+							"nestedkey": "nestedval",
+						},
+					},
+				}
+				return e
+			},
```
djaglowski, line 91: One of the major points of value for this operator is that it allows you to easily move multiple fields at once. Let's have the tests reflect that by adding more fields at each level.

operator/builtin/transformer/flatten/flatten_test.go
```diff
@@ -0,0 +1,159 @@
...
+			func() *entry.Entry {
+				e := newTestEntry()
+				e.Body = map[string]interface{}{
+					"key": "val",
+					"nested": map[string]interface{}{
+						"secondlevel": map[string]interface{}{
+							"nestedkey": "nestedval",
+						},
+					},
+				}
```
djaglowski, line 89: Let's have a test that uses this input but flattens to first level. I want to see that the deeper levels are still intact.

operator/builtin/transformer/flatten/flatten_test.go
```diff
@@ -0,0 +1,159 @@
...
+		e.Timestamp = time.Unix(1586632809, 0)
+		e.Body = map[string]interface{}{
+			"key": "val",
+			"nested": map[string]interface{}{
+				"nestedkey": "nestedval",
+			},
+		}
+		return e
+	}
+	cases := []testCase{
```
djaglowski, line 49: What happens if there is a collision? 
```json
{ 
    "key": "value",
    "obj": {
        "key": "other"
    }
}
```

djaglowski (approved):
Thanks @Mrod1598

PR #78, opened 2021-03-29
Implement new 'Add' operator

Mrod1598:


operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	addOperator := &AddOperator{
+		TransformerOperator: transformerOperator,
+		Field:               c.Field,
+	}
+
+	if reflect.TypeOf(c.Value) == reflect.TypeOf(" ") && strings.Contains(c.Value.(string), "EXPR(") {
```
djaglowski, line 59: We should use [type assertion](https://tour.golang.org/methods/15) instead of reflection.

djaglowski, line 59: Let's make 2 additional changes to validating the string:
1) use `strings.HasPrefix` for the existing check
2) also add a check for `strings.HasSuffix` for the closing paren

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,104 @@
...
+		exprStr := strings.Replace(c.Value.(string), "EXPR(", ``, 1)
```
djaglowski, line 60: `strings.TrimPrefix`

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,104 @@
...
+		if last := len(exprStr) - 1; last >= 0 && exprStr[last] == ')' {
```
djaglowski, line 61: `strings.TrimSuffix`

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,104 @@
...
+			exprStr = exprStr[:last]
+		}
+		compiled, err := expr.Compile(exprStr, expr.AllowUndefinedVariables())
+		if err != nil {
+			return []operator.Operator{addOperator}, fmt.Errorf("failed to compile expression '%s': %w", c.IfExpr, err)
```
djaglowski, line 66: Typically return `nil, error` when there is a failure like this.

In other words, the operator isn't really built in a useful way, so it's no good.

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,104 @@
...
+// Process will process an entry with a restructure transformation.
+func (p *AddOperator) Process(ctx context.Context, entry *entry.Entry) error {
+	return p.ProcessWith(ctx, entry, p.Transform)
+}
+
+// Transform will apply the restructure operations to an entry
+func (p *AddOperator) Transform(entry *entry.Entry) error {
+	if p.Value != nil {
+		return entry.Set(p.Field, p.Value)
+	} else if p.program != nil {
```
djaglowski, line 93: Since the previous `if` block returns, we should not use `else` here

operator/builtin/transformer/add/add_test.go
```diff
@@ -0,0 +1,115 @@
...
+		e.Record = map[string]interface{}{
+			"key": "val",
+			"nested": map[string]interface{}{
+				"nestedkey": "nestedval",
+			},
+		}
+		return e
+	}
+
+	cases := []struct {
```
djaglowski, line 45: Let's have some test cases for adding values to `resource` and `attributes`

djaglowski (changes requested):

docs/operators/add.md
```diff
@@ -0,0 +1,52 @@
...
+| `value`      | required       | The value of the field to be added.
+| `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
+| `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
+
+
+Example usage:
+```yaml
+- type: add
+    field: "key"
+    value: "val"
```
djaglowski, line 28: nit: drop the quotes

operator/builtin/transformer/add/add.go
djaglowski: Bump

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	addOperator := &AddOperator{
+		TransformerOperator: transformerOperator,
+		Field:               c.Field,
+	}
+	_, ok := c.Value.(string)
+	if ok && strings.Contains(c.Value.(string), "EXPR(") {
```
djaglowski, line 59: You can make use of both values that come back from type assertion here:

```
s, ok := c.Value.(string)
if ok && strings.Contains(s, "EXPR(") {
```

We can also make use of `HasPrefix` and `HasSuffix`, so actually these lines should be:
```
s, ok := c.Value.(string)
if ok && strings.HasPrefix(s, "EXPR(") && strings.HasSuffix(s, ")") {
```

operator/builtin/transformer/add/add_test.go
```diff
@@ -0,0 +1,204 @@
...
+			err = remove.Process(context.Background(), tc.input())
+			if tc.expectErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				fake.ExpectEntry(t, tc.output())
+			}
+		})
+	}
+}
```
djaglowski, line 170: This is nice in that it is very "end-to-end", but I think it's clashing with the pattern we've already established in other operators. Let's pull this apart into two sets of tests:
1) Golden configs  (tests unmarshaling from yaml files into config structs)
2) Processing (configure an operator, then validate it processes an entry correctly)

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,102 @@
...
+	return &AddOperatorConfig{
+		TransformerConfig: helper.NewTransformerConfig(operatorID, "add"),
+	}
+}
+
+// AddOperatorConfig is the configuration of a restructure operator
+type AddOperatorConfig struct {
+	helper.TransformerConfig `mapstructure:",squash" yaml:",inline"`
+	Field                    entry.Field `mapstructure:"field" json:"field" yaml:"field"`
+	Value                    interface{} `mapstructure:"value,omitempty" json:"value,omitempty" yaml:"value,omitempty"`
```
djaglowski, line 45: Since `Field` can be an attribute or resource field, we need to define what happens if Value is not a string, and we need tests to ensure that it works as defined.

djaglowski (changes requested):

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+  }
+}
+```
+
+</td>
+</tr>
+</table>
+
+<hr>
+adding nest to body
```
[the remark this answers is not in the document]
djaglowski: Can we add numbers? Arrays?

Is there any valid `json` value that we can't add?

(I think this example suffices to convey the point, but I just want to sanity check that we actually can do this.)

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+  }
+}
+```
+
+</td>
+</tr>
+</table>
+
+<hr>
+adding value to attribute
```
djaglowski, line 156:
```suggestion
Add a value to attributes
```

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+
+```yaml
+- type: add
+    field: $attributes.key2
+    value: 
+      nestedkey: val2
```
djaglowski, line 162: This shouldn't work

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+</td>
+<td>
+
+```json
+{
+  "resource": { },
+  "attributes": {
+     "key2": "val2"
```
djaglowski, line 187: This looks right, but the sample config above is wrong

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+  }
+}
+```
+
+</td>
+</tr>
+</table>
+
+<hr>
+adding value to resource
```
djaglowski, line 200:
```suggestion
Add a value to resource
```

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+
+```yaml
+- type: add
+    field: $resource.key2
+    value: 
+      nestedkey: val2
```
djaglowski, line 206: Wrong config sample

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
...
+    "key1": "val1"
+  }
+}
+```
+
+</td>
+</tr>
+</table>
+
+adding expression to resource
```
djaglowski, line 243:
```suggestion
Add a value to resource using an expression
```

operator/builtin/transformer/add/add.go
djaglowski: So this will just cause an error right?

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,106 @@
...
+		compiled, err := expr.Compile(exprStr, expr.AllowUndefinedVariables())
+		if err != nil {
+			return nil, fmt.Errorf("failed to compile expression '%s': %w", c.IfExpr, err)
+		}
+
+		addOperator.program = compiled
+		return []operator.Operator{addOperator}, nil
+	}
+	addOperator.Value = c.Value
+	return []operator.Operator{addOperator}, nil
```
djaglowski, line 73: Let's extract the expression check and put the simple case first so it can return right away:

```
func isExpr(str string) bool {
	strings.HasPrefix(strVal, "EXPR(") && strings.HasSuffix(strVal, ")")
}

	...
	strVal, ok := c.Value.(string)
	if !ok || !isExpr(strVal) {
		addOperator.Value = c.Value
		return []operator.Operator{addOperator}, nil
	}

	exprStr := strings.TrimPrefix(strVal, "EXPR(")
	exprStr = strings.TrimSuffix(exprStr, ")")

	compiled, err := expr.Compile(exprStr, expr.AllowUndefinedVariables())
	if err != nil {
		return nil, fmt.Errorf("failed to compile expression '%s': %w", c.IfExpr, err)
	}

	addOperator.program = compiled
	return []operator.Operator{addOperator}, nil
```

operator/builtin/transformer/add/add.go
```diff
@@ -0,0 +1,106 @@
...
+		env := helper.GetExprEnv(e)
+		defer helper.PutExprEnv(env)
+
+		result, err := vm.Run(p.program, env)
+		if err != nil {
+			return fmt.Errorf("evaluate value_expr: %s", err)
+		}
+		return e.Set(p.Field, result)
+	}
+	return fmt.Errorf("add: missing required field 'Value'")
```
djaglowski, line 105:
```suggestion
	return fmt.Errorf("add: missing required field 'value'")
```

Uses knows it as `value`, since that's what's in the config

operator/builtin/transformer/add/add_test.go
```diff
@@ -0,0 +1,227 @@
...
+				return e
+			},
+			true,
+		},
+	}
+	for _, tc := range cases {
+		t.Run("BuildandProcess/"+tc.name, func(t *testing.T) {
+			cfgFromMapstructure := tc.op
+			cfgFromMapstructure.OutputIDs = []string{"fake"}
+			cfgFromMapstructure.OnError = "drop"
```
djaglowski, line 209: This isn't unmarshaling the config from yaml. It's just using the maps you defined above.

djaglowski: I see, this is really just a variable naming issue.

The `move` operator also has this problem. Let's fix it in both places on this PR.

Also, please split this file into `config_test.go` and `add_test.go`, to mirror what you did on `move`.

operator/builtin/transformer/add/add_test.go
```diff
@@ -0,0 +1,227 @@
...
+		e.Body = map[string]interface{}{
+			"key": "val",
+			"nested": map[string]interface{}{
+				"nestedkey": "nestedval",
+			},
+		}
+		return e
+	}
+
+	cases := []testCase{
```
djaglowski, line 49: Looks good. Let's add one more test - one that overwrites a value.

operator/builtin/transformer/add/config_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/open-telemetry/opentelemetry-log-collection/entry"
+	"github.com/open-telemetry/opentelemetry-log-collection/operator/helper"
+)
+
+type configTestCase struct {
+	name   string
+	expect *AddOperatorConfig
+}
+
+func TestMoveGoldenConfig(t *testing.T) {
```
djaglowski, line 36:
```suggestion
func TestAddGoldenConfig(t *testing.T) {
```

djaglowski: Or just `TestGoldenConfig`, since it's the only one in this package. If you do this, please also update `move` accordingly.

operator/builtin/transformer/add/add.go
Mrod1598: yeah

docs/operators/add.md
```diff
@@ -0,0 +1,283 @@
+## `add` operator
+
+The `add` operator adds a field to a record. It must have a `field` key and exactly one of `value`.
+
+`field` is a [field](/docs/types/field.md) that will be set to `value` or the evaluated expression.
+
+`value` is a static string that will be added to each entry at the field defined by `field`.
```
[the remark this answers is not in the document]
djaglowski: bump

operator/builtin/transformer/add/add_test.go
```diff
@@ -0,0 +1,270 @@
...
+			}(),
+			newTestEntry,
+			func() *entry.Entry {
+				e := newTestEntry()
+				e.Body = map[string]interface{}{
+					"key": "val",
+					"nested": map[string]interface{}{
+						"nestedkey": "nestedval",
+					},
+					"new": 1,
```
djaglowski, line 241: Since this test is expected to fail, it's odd that the value ended up on the body. Seems like the expected output entry should be the same as the input.

docs/operators/add.md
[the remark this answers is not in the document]
djaglowski: @Mrod1598 This fix is still missing

docs/operators/add.md
```diff
@@ -0,0 +1,281 @@
+## `add` operator
+
+The `add` operator adds a value to an `entry`'s `body`, `attributes`, or `resource`.
+
+`field` is a [field](/docs/types/field.md) that will be set to `value` or the evaluated expression.
+
+`value` is a static string that will be added to each entry at the field defined by `field`.
```
djaglowski, line 7: Since these are defined in the table below, let's remove them. The descriptions don't match up at this point anyways.

djaglowski (approved):

PR #79, opened 2021-03-29
Add Move operator

Mrod1598:


docs/operators/move.md
```diff
@@ -0,0 +1,49 @@
...
+| `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
+| `from`      | required       | The field to move the value out of.   
+| `to`      | required       | The field to move the value into.
+| `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
+| `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
+
+Example usage:
+```yaml
+- type: move
+    from: "key1"
```
djaglowski, line 21: Let's not quote strings in `yaml` unless necessary.

We're a little inconsistent about this across the codebase, but I'd prefer that we default to no quotes going forward.

operator/builtin/transformer/move/move.go
```diff
@@ -0,0 +1,74 @@
...
+	return &MoveOperatorConfig{
+		TransformerConfig: helper.NewTransformerConfig(operatorID, "move"),
+	}
+}
+
+// MoveOperatorConfig is the configuration of a restructure operator
+type MoveOperatorConfig struct {
+	helper.TransformerConfig `mapstructure:",squash" yaml:",inline"`
+	From                     entry.Field
+	To                       entry.Field
```
djaglowski, line 40: These should have `mapstructure`, `json`, and `yaml` tags.

operator/builtin/transformer/move/move.go
```diff
@@ -0,0 +1,74 @@
...
+		TransformerOperator: transformerOperator,
+		From:                c.From,
+		To:                  c.To,
+	}
+
+	return []operator.Operator{addOperator}, nil
+}
+
+type MoveOperator struct {
+	helper.TransformerOperator `mapstructure:",squash" yaml:",inline"`
```
djaglowski, line 59: We only need tags on the config struct.

operator/helper/severity_test.go
```diff
@@ -456,7 +456,7 @@ func TestGoldenSeverityParserConfig(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run("yaml/"+tc.name, func(t *testing.T) {
-			cfgFromYaml, yamlErr := configFromFileViaYaml(path.Join(".", "severitytestdata", fmt.Sprintf("%s.yaml", tc.name)))
+			cfgFromYaml, yamlErr := severityConfigFromFileViaYaml(path.Join(".", "severitytestdata", fmt.Sprintf("%s.yaml", tc.name)))
```
djaglowski, line 459: The severity and time parsing changes appear to be unrelated to the `move` operator. Can you make a separate PR for them?

docs/operators/move.md
```diff
@@ -0,0 +1,49 @@
...
+| `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
+| `from`      | required       | The field to move the value out of.   
+| `to`      | required       | The field to move the value into.
+| `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
+| `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
```
djaglowski, line 16: This doesn't appear to actually exist. I'm ok with dropping it for now, but we may need to add it later.

operator/builtin/transformer/move/move_test.go
```diff
@@ -0,0 +1,166 @@
...
+				e.Record = map[string]interface{}{
+					"key": "val",
+					"NewNested": map[string]interface{}{
+						"nestedkey": "nestedval",
+					},
+				}
+				return e
+			}(),
+		},
+	}
```
djaglowski, line 144: I think we need to cover more cases in these tests. This is a good start, but let's consider more possible field locations, and more types of values. 

Let's consider at least the following:
- `$record.some_string` 
- `$record.some_object` // move the whole object
- `$record.some_object.string` // move nested string
- `$record.some_object.some_object` // move nested object
- `$resource.some_key`
- `$attributes.some_key`

I think we should be able to set up the test so that it runs through all combinations of the above. (It should fail when moving an object to `attributes` or `resource`, but those failures should be tested too).

djaglowski (changes requested):
Looks good overall but there's some extra code and we need some more tests.

operator/builtin/transformer/move/move_test.go
Mrod1598: do you mean that nested should fail when moving to attributes or resource?

PR #80, opened 2021-03-29
Add Remove operator

Mrod1598:

### DIFF
--- a/docs/operators/remove.md
+++ b/docs/operators/remove.md
@@ -10,25 +10,33 @@ It's configured by passing a field to remove.
 | ---        | ---              | ---                                                                                                                                                                                                                                      |
 | `id`       | `restructure`    | A unique identifier for the operator                                                                                                                                                                                                     |
 | `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
-| `field`      | required       | The field to remove.
+| `field`      | required       | The [field](/docs/types/field.md) to remove.
 | `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
 | `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
 
 Example usage:
+
+<hr>
+
+Remove value from body
 ```yaml
 - type: remove 
-    field: "key1"
+    field: 
+    - key1
 ```
 
 <table>
-<tr><td> Input record </td> <td> Output record </td></tr>
+<tr><td> Input Entry </td> <td> Output Entry </td></tr>
 <tr>
 <td>
 
 ```json
 {
-  "key1": "val1",
-  "key2": "val2"
+  "resource": { },
+  "attributes": { },  
+  "body": {
+    "key1": "val1",
+  }
 }
 ```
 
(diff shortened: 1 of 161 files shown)


docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+
+The `remove` operator removes a field from a record.
+
+It's configured by passing a field to remove.
+
+### Configuration Fields
+
+| Field      | Default          | Description                                                                                                                                                                                                                              |
+| ---        | ---              | ---                                                                                                                                                                                                                                      |
+| `id`       | `restructure`    | A unique identifier for the operator                                                                                                                                                                                                     |
```
djaglowski, line 11:
```suggestion
| `id`       | `remove`    | A unique identifier for the operator                                                                                                                                                                                                     |
```

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
+## `remove` operator
+
+The `remove` operator removes a field from a record.
+
+It's configured by passing a field to remove.
```
djaglowski, line 5: Let's just remove this line

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+
+Example usage:
+
+<hr>
+
+Remove value from body
+```yaml
+- type: remove 
+    field: 
+    - key1
```
djaglowski, line 25: Is this right? Does it accept an array?

I think we probably just want one field here. It's easy enough to chain instances of the operator together if needed.

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+}
+```
+
+</td>
+</tr>
+</table>
+
+<hr>
+
+Remove multiple Values from body
```
djaglowski, line 104: Even though `restructure` supports multiple ops at a time, I think part of this decomposition effort should be to simplify as much as possible, which I think means just removing one at a time.

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+```json
+{
+  "resource": { },
+  "attributes": { },  
+  "body": {
+    "object": {
+      "nestedkey": "nestedval"
+    }
+  },
+  "key": "val"
```
djaglowski, line 81: This value should be inside `body`

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+  "resource": { },
+  "attributes": { 
+    "otherkey": "val"
+  },  
+  "body": {
+    "object": {
+      "nestedkey": "nestedval"
+    }, 
+    "key": "val"
+  },
```
djaglowski, line 170: Since the example is focused on attributes, let's keep the body very simple. I think this body distracts from the point of the example.

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+      "nestedkey": "nestedval"
+    }, 
+    "key": "val"
+  }
+}
+```
+
+</td>
+</tr>
+</table>
```
djaglowski, line 243: Let's remove

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+  "resource": { 
+    "otherkey": "val"
+  },
+  "attributes": {  },  
+  "body": {
+    "object": {
+      "nestedkey": "nestedval"
+    }, 
+    "key": "val"
+  },
```
djaglowski, line 270: Same point as above. Keep the body simple since it otherwise draws focus away from the example

docs/operators/remove.md
```diff
@@ -0,0 +1,343 @@
...
+    "object": {
+      "nestedkey": "nestedval"
+    }, 
+    "key": "val"
+  }
+}
+```
+
+</td>
+</tr>
```
djaglowski, line 342: Let's remove

operator/builtin/transformer/remove/config_test.go
```diff
@@ -0,0 +1,144 @@
+package remove
```
djaglowski, line 1: I think the package typically goes under the copyright

operator/builtin/transformer/remove/remove.go
```diff
@@ -0,0 +1,82 @@
...
+	return &RemoveOperatorConfig{
+		TransformerConfig: helper.NewTransformerConfig(operatorID, "remove"),
+	}
+}
+
+// RemoveOperatorConfig is the configuration of a restructure operator
+type RemoveOperatorConfig struct {
+	helper.TransformerConfig `mapstructure:",squash" yaml:",inline"`
+
+	Fields []entry.Field `mapstructure:"fields"  json:"fields" yaml:"fields"`
```
djaglowski, line 41: Let's just accept one

operator/builtin/transformer/remove/remove.go
```diff
@@ -0,0 +1,82 @@
...
+	if p.Fields != nil {
+		for _, field := range p.Fields {
+			_, exist := entry.Delete(field)
+			if !exist {
+				return fmt.Errorf("remove: field does not exist")
+			}
+		}
+		return nil
+	}
+	return fmt.Errorf("remove: missing required field 'fields'")
```
djaglowski, line 81: This check needs to happen in `Build`. We can know whether it is set at that point and error out then.

There are two problems with doing the check here:
1) The user doesn't get immediate feedback about their configuration error
2) The check has to be performed for every single log entry, so becomes an unnecessary performance hit

operator/builtin/transformer/remove/remove_test.go
```diff
@@ -0,0 +1,213 @@
...
+					"nested": map[string]interface{}{
+						"nestedkey": "nestedval",
+					},
+				}
+				return e
+			},
+			false,
+		},
+		{
+			"remove_obj",
```
djaglowski, line 70: Let's have a test that removes `nested.nestedKey`, so that we're validation the drilling aspect of this.

djaglowski (changes requested):
A couple important points on this one:
1) I don't think we should support multi-remove. (We don't support multi-add, multi-move, etc)
2) The config needs to be validated in the `Build` method, rather than `Transform`

djaglowski (commented):
Just need to standardize the descriptions in the doc, but otherwise looks good.

djaglowski (approved):
LGTM

PR #81, opened 2021-03-29
Add Retain operator

Mrod1598:

### DIFF
--- a/docs/operators/retain.md
+++ b/docs/operators/retain.md
@@ -14,7 +14,7 @@ The `retain` operator keeps the specified list of fields, and removes the rest.
 
 Example usage:
 <hr>
-retain fields in the body
+Retain fields in the body
 
 ```yaml
 - type: retain
@@ -60,7 +60,7 @@ retain fields in the body
 </table>
 
 <hr>
-retain an object in the body
+Retain an object in the body
 
 ```yaml
 - type: retain
@@ -106,7 +106,7 @@ retain an object in the body
 </table>
 
 <hr>
-retain fields from resource
+Retain fields from resource
 
 ```yaml
 - type: retain
@@ -155,7 +155,7 @@ retain fields from resource
 </table>
 
 <hr>
-retain fields from attributes
+Retain fields from attributes
 
 ```yaml
 - type: retain
@@ -202,7 +202,7 @@ retain fields from attributes
 </table>
 
 <hr>
-retain fields from all sources
+Retain fields from all sources
 
 ```yaml
 - type: retain
--- a/docs/operators/retain.md
+++ b/docs/operators/retain.md
@@ -8,13 +8,13 @@ The `retain` operator keeps the specified list of fields, and removes the rest.
 | ---        | ---              | ---                                                                                                                                                                                                                                      |
 | `id`       | `retain`    | A unique identifier for the operator                                                                                                                                                                                                     |
 | `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
-| `fields`      | required         | A list of fields to be kept.                                                                                                                                                     |
+| `fields`      | required         | A list of [fields](/docs/types/field.md)  to be kept.                                                                                                                                                     |
 | `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
 | `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
 
 Example usage:
 <hr>
-retain body fields
+retain fields in the body
 
 ```yaml
 - type: retain
@@ -60,7 +60,7 @@ retain body fields
 </table>
 
 <hr>
-retain body object
+retain an object in the body
 
 ```yaml
 - type: retain
@@ -106,7 +106,7 @@ retain body object
 </table>
 
 <hr>
-retain resource fields
+retain the resource fields
 
 ```yaml
 - type: retain
@@ -155,7 +155,7 @@ retain resource fields
 </table>
 
 <hr>
-retain attributes fields
+retain the attributes fields
 
 ```yaml
 - type: retain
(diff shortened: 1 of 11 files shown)


docs/operators/retain.md
```diff
@@ -0,0 +1,256 @@
...
+| `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
+| `fields`      | required         | A list of [fields](/docs/types/field.md)  to be kept.                                                                                                                                                     |
+| `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
+| `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
+
+Example usage:
+<hr>
+retain fields in the body
```
djaglowski, line 17: Capitalize

docs/operators/retain.md
```diff
@@ -0,0 +1,256 @@
...
+
+<hr>
+retain fields from all sources
+
+```yaml
+- type: retain
+    fields:
+      - $resource.key1
+      - $attributes.key3
+      - key5
```
djaglowski, line 212: I think that by supporting `resource` and `attributes`, we've actually introduced a bit of a design problem for this operator. 

Basically, I think it's generally unlikely that users will want to drop resource and attribute fields, so the primary use case is probably just for body.

I need to think on this one a bit.

Mrod1598: we could do something like this 
`
// Transform will apply the retain operation to an entry
func (p *RetainOperator) Transform(e *entry.Entry) error {
	newEntry := entry.New()
	containsRorA := false
	for _, field := range p.Fields {
		if reflect.TypeOf(field) == reflect.TypeOf(entry.BodyField{}) || reflect.TypeOf(field) == reflect.TypeOf(entry.ResourceField{}) {
			containsRorA = true
		}
	}
	if containsRorA {
		newEntry.Timestamp = e.Timestamp
		for _, field := range p.Fields {
			val, ok := e.Get(field)
			if !ok {
				return fmt.Errorf("retain: field does not exist: %s", field)
			}
			err := newEntry.Set(field, val)
			if err != nil {
				return err
			}
		}
		*e = *newEntry
		return nil
	}
	newEntry.Timestamp = e.Timestamp
	newEntry.Attributes = e.Attributes
	newEntry.Resource = e.Resource
	for _, field := range p.Fields {
		val, ok := e.Get(field)
		if !ok {
			return fmt.Errorf("retain: field does not exist: %s", field)
		}
		err := newEntry.Set(field, val)
		if err != nil {
			return err
		}
	}
	*e = *newEntry
	return nil
}
`
that way if they on't specify any attributes or resources we will just keep everything

docs/operators/retain.md
```diff
@@ -0,0 +1,261 @@
...
+| `output`   | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
+| `fields`      | required         | A list of [fields](/docs/types/field.md)  to be kept.                                                                                                                                                     |
+| `on_error` | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md)                                                                                                                                          |
+| `if`       |                  | An [expression](/docs/types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
+<hr>
+<b>NOTE:</b> If no fields in a group (attributes,resource, or body) are specified, that entire group will be retained.
```
djaglowski, line 15:
```suggestion
<b>NOTE:</b> If no fields in a group (attributes, resource, or body) are specified, that entire group will be retained.
```

operator/builtin/transformer/retain/retain.go
```diff
@@ -0,0 +1,116 @@
...
+	Fields                   []entry.Field `mapstructure:"fields" json:"fields" yaml:"fields"`
+}
+
+// Build will build a retain operator from the supplied configuration
+func (c RetainOperatorConfig) Build(context operator.BuildContext) ([]operator.Operator, error) {
+	transformerOperator, err := c.TransformerConfig.Build(context)
+	if err != nil {
+		return nil, err
+	}
+	if c.Fields == nil {
```
djaglowski, line 50: It's probably also possible that `len(c.Fields) == 0`, and if so, I believe we should handle it here as well.

operator/builtin/transformer/retain/retain.go
```diff
@@ -0,0 +1,116 @@
...
+	return []operator.Operator{retainOp}, nil
+}
+
+// RetainOperator keeps the given fields and deletes the rest.
+type RetainOperator struct {
+	helper.TransformerOperator
+	Fields                  []entry.Field
+	ContainsBodyFields      bool
+	ContainsAttributeFields bool
+	ContainsResourceFields  bool
```
djaglowski, line 80: Just a nit, but in terms of naming, I think `AllBodyFields` would be better.

The current names are very clear when you are thinking about them in terms of what the configuration says, but `AllResourceFields` is a bit more clear in terms of what should be retained, which is how the variables are actually used.

operator/builtin/transformer/retain/retain.go
```diff
@@ -0,0 +1,116 @@
...
+func (p *RetainOperator) Transform(e *entry.Entry) error {
+	newEntry := entry.New()
+	newEntry.Timestamp = e.Timestamp
+
+	if !p.ContainsAttributeFields {
+		newEntry.Attributes = e.Attributes
+	}
+	if !p.ContainsResourceFields {
+		newEntry.Resource = e.Resource
+	}
```
djaglowski, line 98: Super nit, but would prefer that handling of resource and attributes are in the same order as `Build`.

operator/builtin/transformer/retain/retain.go
```diff
@@ -0,0 +1,116 @@
...
+	if !p.ContainsBodyFields {
+		newEntry.Body = e.Body
+	}
```
djaglowski, line 101: Nice job keeping this minimal impact per-entry.

operator/builtin/transformer/retain/retain_test.go
```diff
@@ -0,0 +1,322 @@
...
+		e.Body = map[string]interface{}{
+			"key": "val",
+			"nested": map[string]interface{}{
+				"nestedkey": "nestedval",
+			},
+		}
+		return e
+	}
+
+	cases := []testCase{
```
djaglowski, line 50: Looks good. Let's have a test that retains a non-existant field, to validate that it behaves as expected.

djaglowski (commented):
Looks great. Just a couple nits

operator/builtin/transformer/retain/retain_test.go
Mrod1598: So, if you give a body field that doesn't actually exist it still treats it as a body field and just retains nothing in the body field and deletes everything. is that how we want it to work? @djaglowski. 

the config and the entry that results from it, if you were curious
```
	{
			"retain_a_non_existant_key",
			false,
			func() *RetainOperatorConfig {
				cfg := defaultCfg()
				cfg.Fields = append(cfg.Fields, entry.NewResourceField("aNonExsistentKey"))
				return cfg
			}(),
			newTestEntry,
			func() *entry.Entry {
				e := newTestEntry()
				return e
			},
		},
```

djaglowski: That behavior seems fine to me. Having the test will still be good so that we know if the behavior unexpectedly changes in the future.

djaglowski (approved):
Thanks @Mrod1598

PR #89, opened 2021-04-05
Remove database, pass persistor to Start methods

djaglowski:
This PR prepares the library for the use of the storage extension framework, which is established [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2883).

The idea is that state will be persisted at a higher level, and only a simple persistence interface is necessary within this library.

- Removes the concept of a database, and the associated dependency on bbolt
- Reworks the `persister` interface, removing `Sync` and `Load`, but adding `context.Context` to all methods
- Passes a `persister` to each `Start` method (Agent, Pipeline, Operator)
- Moves `persister` into the `operator` package, because it is similar to `BuildContext` in that it is a sort of `RunContext`.


djaglowski:
@jsirianni Would you mind giving this a preliminary review when you have a chance? Please note that the PR is requesting to merge to a non-master branch. The idea here is to make this code available for non-local testing (CI on the contrib repo, for example)

jsirianni (approved):

PR #90, opened 2021-04-05
Support whole number float64 type for severity parsing

jsirianni:
Severity parsing should support float64 values when they are whole numbers. Golang JSON parsing will consider any number a float64 unless it is unmarshalled to a struct using specific types (int, int64, float64, etc).

Trying to parse the `pri` field as a Severity will result in an "unsupported type float64" error.
```json
{
  "uid": 62,
  "time": 1617410898,
  "pri": 6,
  "tag": "pve-ha-lrm",
  "pid": 33690,
  "node": "pve-1",
  "user": "root@pam",
  "msg": "end task UPID:pve-1:0000839B:000C5C3F:6067BB15:qmshutdown:100:root@pam: OK"
}
```

Additional context:
- Stanza PR https://github.com/observIQ/stanza/pull/268
- Stanza Issue https://github.com/observIQ/stanza/issues/267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Added float64 to Severity parser's supported types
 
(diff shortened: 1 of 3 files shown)


djaglowski:
> when they are whole numbers

Just to clarify, this works for non-whole numbers as well, right? The number will just lose its decimal precision?

djaglowski (approved):
LGTM

Issue #95, opened 2021-04-12
entry/entry.go:44:12: struct with 136 pointer bytes could be 128


pmalek-sumo:
[fieldalignment](https://pkg.go.dev/golang.org/x/tools@v0.1.0/go/analysis/passes/fieldalignment/cmd/fieldalignment) report that (among others) `entry.Entry` be made smaller by reordering its fields:

```
struct with 136 pointer bytes could be 128
```

PR #96, opened 2021-04-12
Make entry.Entry smaller by reordeing fields

pmalek-sumo:
Changes suggest by [fieldalignment](https://pkg.go.dev/golang.org/x/tools@v0.1.0/go/analysis/passes/fieldalignment/cmd/fieldalignment)

```
struct with 136 pointer bytes could be 128
```

Fixes: https://github.com/open-telemetry/opentelemetry-log-collection/issues/95


djaglowski (approved):
Nice, thanks @pmalek-sumo!

PR #97, opened 2021-04-12
Add double-star mutli-level directory glob matching to filelog receiver

pmalek-sumo:
Closes: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2329

### DIFF
--- a/operator/builtin/input/file/config_test.go
+++ b/operator/builtin/input/file/config_test.go
@@ -81,6 +81,33 @@ func TestConfig(t *testing.T) {
 			}(),
 		},
 		{
+			"include_glob_double_asterisk",
+			false,
+			func() *InputConfig {
+				cfg := defaultCfg()
+				cfg.Include = append(cfg.Include, "**.log")
+				return cfg
+			}(),
+		},
+		{
+			"include_glob_double_asterisk_nested",
+			false,
+			func() *InputConfig {
+				cfg := defaultCfg()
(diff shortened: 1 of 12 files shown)


operator/builtin/input/file/config_test.go
```diff
@@ -80,6 +80,24 @@ func TestConfig(t *testing.T) {
...
+				cfg.Include = append(cfg.Include, "**.log")
+				return cfg
+			}(),
+		},
+		{
+			"include_glob_double_asterisk_nested",
+			false,
+			func() *InputConfig {
+				cfg := defaultCfg()
+				cfg.Include = append(cfg.Include, "directory/**/*.log")
```
djaglowski, line 97: Could we add one more test for `**/directory/**/*.log`?

djaglowski (commented):
Looks good, just a couple minor suggestions.

Could you also update the `file_input` documentation so that users can easily find the exact rules that govern paths. It's ok to directly reference the [doublestar docs](https://github.com/bmatcuk/doublestar#patterns). This is done elsewhere, such as in the [expressions](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/types/expression.md) doc.

operator/builtin/input/file/config_test.go
pmalek-sumo: Done

djaglowski (approved):
Thanks @pmalek-sumo. LGTM

PR #101, opened 2021-04-16
Upgrade otel collector core dependency

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


tigrannajaryan:
@djaglowski I did not realize we have this dependency. Given that we plan to move `filelog` to the core we are going to have a circular dependency. We need to think about how to break this.

tigrannajaryan (approved):

PR #112, opened 2021-04-20
Update Remove operator to support removing the entire body, resource or attributes

Mrod1598:
To support deleting the entire object, A new field struct was needed so we can catch whether any of the objects are specified. Also, to enable this, some fields were needed to face outwards from the field package that aren't currently. Those fields were then updated throughout the project to reflect that.


operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+		*f = RootableField{allBody: true}
+		return nil
+	}
+
+	field, err := entry.FieldFromString(s)
+	if err != nil {
+		return err
+	}
+	*f = RootableField{Field: &field}
+	return nil
```
djaglowski, line 64: Since this is exactly the same in both unmarshal methods, let's extract it

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/open-telemetry/opentelemetry-log-collection/entry"
+)
+
+// RootableField represents a potential field on an entry.
+// It differs from a normal Field in that it allows users to
+// specify `$resource` or `$attributes` with the intention
+// of referring to "all" fields within those groups.
+// It is used to get, set, and delete values at this field.
+// It is deserialized from JSON dot notation.
+type RootableField struct {
```
djaglowski, line 29: I think we should probably make this type private, since it has some special behaviors that only make sense for the `remove` operator. 

We may eventually need to generalize it, but for now, let's not imply that it's ready for use outside of this operator.

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+	*entry.Field
+	allResource   bool
+	allAttributes bool
+	allBody       bool
```
djaglowski, line 33: I could be wrong, but I think `$body` is already supported in the underlying `entry.Field`, so we probably do not need any of the special handling for it.

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if s == entry.AttributesPrefix {
+		*f = RootableField{allAttributes: true}
+		return nil
+	}
+
+	if s == entry.BodyPrefix {
+		*f = RootableField{allBody: true}
+		return nil
+	}
```
djaglowski, line 57: Probably can remove special handling for `$body`

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
```
djaglowski, line 88: Probably can remove special handling for `$body`

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+	return f.Field.Set(entry, value)
+}
+
+func (f *RootableField) Delete(entry *entry.Entry) (interface{}, bool) {
+	if f.allAttributes || f.allResource {
+		return nil, false
+	}
+	return f.Field.Delete(entry)
+}
```
djaglowski, line 117: I'm not sure these methods are doing anything useful.

The operator only makes use of the `Delete` method, and only when `$attributes` or `$resource` is specified. When anything else is specified, the behavior is the same as the underlying `entry.Field`.

operator/builtin/transformer/remove/remove.go
```diff
@@ -47,7 +47,8 @@ func (c RemoveOperatorConfig) Build(context operator.BuildContext) ([]operator.O
 	if err != nil {
 		return nil, err
 	}
-	if c.Field == entry.NewNilField() {
+
+	if c.Field == NewNilRootableField() {
```
djaglowski, line 51:
```suggestion
	if c.Field.Field == entry.NewNilField() {
```

I think this should work, and then we don't need `NewNilRootableField()`

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+
+func NewResourceField(key string) RootableField {
+	field := entry.NewResourceField(key)
+	return RootableField{Field: &field}
+}
+
+func NewAttributeField(key string) RootableField {
+	field := entry.NewAttributeField(key)
+	return RootableField{Field: &field}
+}
```
djaglowski, line 147: I think these can be private, or maybe even moved to test files.

operator/builtin/transformer/remove/rootable_field.go
```diff
@@ -0,0 +1,147 @@
...
+	if f.allResource {
+		return entry.ResourcePrefix
+	}
+	return f.Field.String()
+}
+
+func NewNilRootableField() RootableField {
+	nilField := entry.NewNilField()
+	return RootableField{Field: &nilField}
+}
```
djaglowski, line 132: Can likely be removed. See other comment where this is used.

entry/field.go
```diff
@@ -58,28 +59,28 @@ func (f *Field) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err != nil {
 		return err
 	}
-	*f, err = fieldFromString(s)
+	*f, err = FieldFromString(s)
 	return err
 }
 
-func fieldFromString(s string) (Field, error) {
+func FieldFromString(s string) (Field, error) {
```
djaglowski, line 66: Since this is now exported, I think we should probably just call it `NewField`.

djaglowski (commented):
This looks pretty good, but lots of feedback.

In addition to what's pointed out in the code, please make sure to add comments to all exported types and funcs.

operator/builtin/transformer/remove/rootable_field.go
djaglowski: bump

operator/builtin/transformer/remove/rootable_field.go
djaglowski: bump

djaglowski:
@Mrod1598 Please fix build issues.

operator/builtin/transformer/remove/rootable_field.go
djaglowski: @Mrod1598 This may have gotten buried.

operator/builtin/transformer/remove/rootable_field.go
djaglowski: @Mrod1598 This may have gotten buried.

djaglowski (approved):
Nice work @Mrod1598

PR #125, opened 2021-05-04
[WIP] add multiline support to TCP

sumo-drosiek:
This is my proposal for multiline support in TCP
It's not finished yet, but please see if this is acceptable direction

 - moving encoding and multiline to helper to reuse it in TCP
 - create reader for TCP input operator

### DIFF
--- a/operator/builtin/input/file/config.go
+++ b/operator/builtin/input/file/config.go
@@ -112,7 +112,7 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator, 
 		return nil, err
 	}
 
-	multiline, err := c.Multiline.Build(context, encoding, false)
+	splitFunc, err := c.Multiline.Build(context, encoding, false)
 	if err != nil {
 		return nil, err
 	}
@@ -141,7 +141,7 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator, 
 		InputOperator:      inputOperator,
 		Include:            c.Include,
 		Exclude:            c.Exclude,
-		SplitFunc:          multiline.SplitFunc,
+		SplitFunc:          splitFunc,
 		PollInterval:       c.PollInterval.Raw(),
 		FilePathField:      filePathField,
 		FileNameField:      fileNameField,
--- a/operator/builtin/input/file/config.go
+++ b/operator/builtin/input/file/config.go
@@ -112,7 +112,7 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator, 
 		return nil, err
 	}
 
-	multiline, err := c.Multiline.Build(context, encoding)
+	multiline, err := c.Multiline.Build(context, encoding, false)
 	if err != nil {
 		return nil, err
 	}
--- a/operator/builtin/input/file/config.go
+++ b/operator/builtin/input/file/config.go
@@ -16,13 +16,9 @@ package file
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/bmatcuk/doublestar/v3"
-	"golang.org/x/text/encoding"
-	"golang.org/x/text/encoding/ianaindex"
-	"golang.org/x/text/encoding/unicode"
 
 	"github.com/open-telemetry/opentelemetry-log-collection/entry"
 	"github.com/open-telemetry/opentelemetry-log-collection/operator"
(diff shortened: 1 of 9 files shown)


sumo-drosiek:
Sorry for the quality, but I want to get feedback of general idea quickly to resume work on that Tomorrow

djaglowski:
@sumo-drosiek, Thanks for taking a crack at this. 

I agree with moving encoding and multiline to helper. I may have some minor comments later, but overall this looks great.

However, I don't see the value of the Reader. Can we just enhance `goHandleMessages` to make use of encoding and multiline? Maybe I am missing something, but this seems like an unnecessary abstraction.

operator/builtin/input/file/config.go
```diff
@@ -153,7 +141,7 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator,
 		InputOperator:      inputOperator,
 		Include:            c.Include,
 		Exclude:            c.Exclude,
-		SplitFunc:          splitFunc,
+		SplitFunc:          multiline.SplitFunc,
```
sumo-drosiek, line 144: Would it be better to return `SplitFunc` by `multiline.Build`?

operator/builtin/input/tcp/tcp.go
```diff
@@ -49,17 +50,21 @@ func init() {
...
 
 // TCPInputConfig is the configuration of a tcp input operator.
 type TCPInputConfig struct {
 	helper.InputConfig `yaml:",inline"`
 
-	MaxBufferSize helper.ByteSize         `json:"max_buffer_size,omitempty" yaml:"max_buffer_size,omitempty"`
-	ListenAddress string                  `json:"listen_address,omitempty" yaml:"listen_address,omitempty"`
-	TLS           *helper.TLSServerConfig `json:"tls,omitempty" yaml:"tls,omitempty"`
-	AddAttributes bool                    `json:"add_attributes,omitempty" yaml:"add_attributes,omitempty"`
+	MaxBufferSize helper.ByteSize         `mapstructure:"max_log_size,omitempty"          json:"max_log_size,omitempty"         yaml:"max_log_size,omitempty"`
```
djaglowski, line 62: nit: For the sake of readability, should we change the field name to `MaxLogSize`?

operator/builtin/input/file/config.go
djaglowski: Seems ok either way, as long as the error can be surfaced or handled appropriately.

djaglowski (approved):
This looks great. Thanks @sumo-drosiek!

sumo-drosiek:
I change multiline to return splitFunc directly and renamed MaxBufferSize to MaxLogSize

sumo-drosiek:
ready to go :tada:

sumo-drosiek:
@djaglowski I noticed that I use encoder only during splitting messages. I need to use it in decoding messages as well. Due to that encoder helper will be probably changed (extended with decode method)

operator/helper/encoding.go
```diff
@@ -38,8 +39,37 @@ type EncodingConfig struct {
...
+}
+
+type Encoding struct {
+	Encoding encoding.Encoding
+}
+
+// decode converts the bytes in msgBuf to utf-8 from the configured encoding
+func (e *Encoding) Decode(msgBuf []byte) (string, error) {
+	decodeBuffer := make([]byte, 1<<12)
+	decoder := e.Encoding.NewDecoder()
```
sumo-drosiek, line 60: I create it every call to avoid race conditions
I left file reader implementation in file operator because it's optimized for that usecase

sumo-drosiek:
@djaglowski Please see latest commit and I think we are good to go :)

operator/helper/encoding.go
```diff
@@ -38,8 +39,37 @@ type EncodingConfig struct {
...
+		decoder.Reset()
+		nDst, _, err := decoder.Transform(decodeBuffer, msgBuf, true)
+		if err != nil && err == transform.ErrShortDst {
+			decodeBuffer = make([]byte, len(decodeBuffer)*2)
+			continue
+		} else if err != nil {
+			return "", fmt.Errorf("transform encoding: %s", err)
+		}
+		return string(decodeBuffer[:nDst]), nil
+	}
```
djaglowski, line 72: Nit, but we may want to invert the error checking to flatten the code:
```go
if err == nil {
	return string(decodeBuffer[:nDst]), nil
}
if err == transform.ErrShortDst {
	decodeBuffer = make([]byte, len(decodeBuffer)*2)
	continue
} 
return "", fmt.Errorf("transform encoding: %s", err)
```

operator/helper/encoding.go
djaglowski: Seems reasonable. If it proves to be a performance bottleneck at some point, we can look at optimizing then.

djaglowski (approved):
LGTM

PR #126, opened 2021-05-07
Add fieldalignment make target

### LINKED ISSUE #98: Add fieldalignment make target
See #95

It would be nice to have this functionality readily available so we can easily identify further improvements, and also so that we can easily validate that we have not regressed.

chasekiefer:
resolves #98


chasekiefer:
@djaglowski Thank you for checking this out. 

This problem is actually something I noticed while doing testing. It runs fine in my IDE terminal, but in a zsh shell I needed to set the ENV vars GOPATH, GOROOT, and PATH. 

I believe it should work if you run:
```
export PATH=$PATH:$GOPATH/bin
```

djaglowski:
@chasekiefer Unfortunately, that does not appear to be the problem. `$GOPATH/bin` is in my `PATH`, and as I noted:

> `fieldalignment` is not present in `$GOPATH/bin` after running `make install-tools`

PR #127, opened 2021-05-10
Add encoding, max_log_size and multiline to udp

sumo-drosiek:


operator/builtin/input/udp/udp.go
```diff
@@ -53,6 +69,12 @@ func (c UDPInputConfig) Build(context operator.BuildContext) ([]operator.Operato
 		return nil, err
 	}
 
+	// If MaxLogSize not set, set sane default in order to remain
+	// backwards compatible with existing plugins and configurations
+	if c.MaxLogSize == 0 {
+		c.MaxLogSize = DefaultMaxLogSize
+	}
+
```
djaglowski, line 77: `tcp_input` does a check against `minBufferSize`. We should probably do the same here. Also, since we renamed `MaxBufferSize` to `MaxLogSize`, we should probably call this `minLogSize`.

```go
if c.MaxLogSize < minBufferSize {
	return nil, fmt.Errorf("invalid value for parameter 'max_log_size', must be equal to or greater than %d bytes", minBufferSize)
}
```

operator/builtin/input/udp/udp.go
```diff
@@ -62,11 +84,24 @@ func (c UDPInputConfig) Build(context operator.BuildContext) ([]operator.Operato
...
+	splitFunc, err := c.Multiline.Build(context, encoding.Encoding, true)
+	if err != nil {
+		return nil, err
+	}
+
 	udpInput := &UDPInput{
 		InputOperator: inputOperator,
 		address:       address,
-		buffer:        make([]byte, 8192),
+		buffer:        make([]byte, DefaultMaxLogSize),
```
djaglowski, line 100: Should this be `buffer:        make([]byte, int(c.MaxLogSize))`?

djaglowski: Or if the point is to be able to buffer multiple logs, then perhaps use a multiple of `c.MaxLogSize`? (And maybe this becomes a configurable value later?)

djaglowski (commented):
Looks good. Just a couple minor questions.

operator/builtin/input/udp/udp.go
sumo-drosiek: This is tricky as we should read full UDP packet at once, since rest of data is going to be loss:
https://stackoverflow.com/a/24705791

I will set it to 64K as it is maximum possible udp packet size

sumo-drosiek:
@djaglowski I unified TCP and UDP buffering

docs/operators/udp_input.md
```diff
@@ -4,27 +4,56 @@ The `udp_input` operator listens for logs from UDP packets.
...
+| `output`          | Next in pipeline | The connected operator(s) that will receive all outbound entries                                                   |
+| `listen_address`  | required         | A listen address of the form `<ip>:<port>`                                                                         |
+| `write_to`        | $                | The body [field](/docs/types/field.md) written to when creating a new log entry                                    |
+| `attributes`      | {}               | A map of `key: value` pairs to add to the entry's attributes                                                       |
+| `resource`        | {}               | A map of `key: value` pairs to add to the entry's resource                                                         |
+| `add_attributes`  | false            | Adds `net.transport`, `net.peer.ip`, `net.peer.port`, `net.host.ip` and `net.host.port` attributes                 |
+| `max_log_size`    | `1MiB`           | The maximum size of a log entry to read before failing. Protects against reading large amounts of data into memory |
```
djaglowski, line 16: Since UDP sets a maximum size, and since we are defaulting to that size, I think parameter is only serving as a memory tuning knob. Is a worthwhile use case? Is there another use case I'm missing?

If we do wish to keep it, then we need to reflect the correct default value, and change the description to explain what it can be used for.

operator/builtin/input/tcp/tcp.go
```diff
@@ -32,9 +32,9 @@ import (
 )
 
 const (
-	// minBufferSize is the initial size used for buffering
+	// minLogSize is the initial size used for buffering
```
djaglowski, line 35: I believe this comment is inaccurate

operator/builtin/input/tcp/tcp.go
```diff
@@ -32,9 +32,9 @@ import (
...
 	// TCP input
-	minBufferSize = 64 * 1024
+	minLogSize = 64 * 1024
```
djaglowski, line 37: Nit: I know I just suggested this name, but it's actually quite difficult to reason about its actual purpose here. 

The value actually represents "the minimum allowable maximum log size", so maybe `minMaxLogSize`? This is not great either, but at least makes sense next to `DefaultMaxLogSize`. 

I'm open to suggestions here, or we can just leave it (as long as the comment is clearer)

docs/operators/udp_input.md
sumo-drosiek: not sure. Does max_log_size for other operators have other purpose than memory tune?

operator/builtin/input/udp/udp.go
```diff
@@ -27,6 +29,20 @@ import (
 	"github.com/open-telemetry/opentelemetry-log-collection/operator/helper"
 )
 
+const (
+	// Maximum UDP packet size
+	MaxUDPSize = 64 * 1024
+
+	// minLogSize is the initial size used for buffering
+	// TCP input
```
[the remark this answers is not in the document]
sumo-drosiek: should be better now

operator/builtin/input/tcp/tcp.go
sumo-drosiek: naming is not my best skill, but to keep convention I changed it to `minMaxLogSize`

docs/operators/udp_input.md
djaglowski: I think that in the contexts of `tcp_input` and `file_input`, this setting works as described, in that it can "Protect against reading large amounts of data into memory". 

In any case, I see that you've removed it from the doc, but it is still present throughout the code. We can remove related values such as `minMaxLogSize` as well, since they only served to validate this setting.

operator/builtin/input/udp/udp.go
```diff
@@ -35,15 +42,23 @@ func init() {
...
 
 // UDPInputConfig is the configuration of a udp input operator.
 type UDPInputConfig struct {
 	helper.InputConfig `yaml:",inline"`
 
-	ListenAddress string `json:"listen_address,omitempty" yaml:"listen_address,omitempty"`
-	AddAttributes bool   `json:"add_attributes,omitempty" yaml:"add_attributes,omitempty"`
+	ListenAddress string                 `mapstructure:"listen_address,omitempty"        json:"listen_address,omitempty"       yaml:"listen_address,omitempty"`
+	AddAttributes bool                   `mapstructure:"add_attributes,omitempty"        json:"add_attributes,omitempty"       yaml:"add_attributes,omitempty"`
+	MaxLogSize    helper.ByteSize        `mapstructure:"max_log_size,omitempty"          json:"max_log_size,omitempty"         yaml:"max_log_size,omitempty"`
```
djaglowski, line 59:
```suggestion
```

operator/builtin/input/tcp/tcp.go
```diff
@@ -79,8 +79,8 @@ func (c TCPInputConfig) Build(context operator.BuildContext) ([]operator.Operato
 		c.MaxLogSize = DefaultMaxLogSize
 	}
 
-	if c.MaxLogSize < minBufferSize {
-		return nil, fmt.Errorf("invalid value for parameter 'max_log_size', must be equal to or greater than %d bytes", minBufferSize)
+	if c.MaxLogSize < minMaxLogSize {
+		return nil, fmt.Errorf("invalid value for parameter 'max_log_size', must be equal to or greater than %d bytes", minMaxLogSize)
```
djaglowski, line 83:
```suggestion
```

docs/operators/udp_input.md
djaglowski: I think that in the contexts of `tcp_input` and `file_input`, this setting works as described, in that it can "Protect against reading large amounts of data into memory". 

In any case, I see that you've removed it from the udp.

operator/builtin/input/tcp/tcp.go
djaglowski: This is a weird one to try to name. Thanks for humoring my nit picking.

djaglowski (approved):
Looks good. Just one extraneous usage of max log size in the UDP Config

sumo-drosiek:
@djaglowski All issues fixed :)

PR #132, opened 2021-05-11
Add SetID func to operator Configs

Mrod1598:
This is to give the ability to modify the ids for autogeneration of ids when there are multiple ops with the same name.


djaglowski:
I would like to see a complete implementation of using this method before we accept. Closing for now.

PR #139, opened 2021-05-14
Plugin ID testing

Mrod1598:
Adds testing for plugins to ensure IDs are generated as expected within a plugin. This is also in preparation for the Autogened ID functionality to be added soon. 

Also, this changes the way Default outputs are set due to plugins being mishandled when they're the next operator in line in the pipeline.


plugin/config_test.go
```diff
@@ -133,3 +133,59 @@ pipeline:
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "reached max plugin depth")
 }
+
+func TestPluginIDs(t *testing.T) {
```
Mrod1598, line 137: this had to be in the plugin package because it created an import cycle in the pipeline package. Above this, there is a buildoperators call.

plugin/config_test.go
```diff
@@ -101,6 +101,149 @@ output: stdout
 	require.Equal(t, "noop", noop.OperatorType)
 }
 
+type PluginIDTestCase struct {
```
djaglowski, line 104: Aside from changes in this file, are the other changes necessary to make these tests pass? If so, what functionality do they change?

PR #140, opened 2021-05-15
fix data loss between logrotate

### LINKED ISSUE #85: file_input - Data loss when tailing symlink file with log rotation (k8s container logs)
See [contrib issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2946) for details.

rockb1017:
fixes #85 
Main thing I needed to do is to keep track of files it is tailing and keep them opened through polling cycle. 
To do that, i made much changes to the poll method. 

I am having issues with FingerPrint.FirstBytes. test sometimes passes and sometimes fails. 
Please review and suggest changes I can make.


tigrannajaryan:
Please also add a new test that verifies the bug fix. The test presumably should fail with the implementation before the fix and should pass after the fix.

djaglowski:
In an effort to gain confidence in fingerprint comparison logic, I added another test here in #145.

I have yet to see a case where `fingerprint.StartsWith` fails when it should pass, so I'm strongly suspecting that this is not the problem. If the problem is with this method, we should be able to reproduce it in a test.

I strongly suspect that the issue is not directly with fingerprint comparison.

Tomorrow I will get my head further into this problem.

rockb1017:
posting some details from what I see with failing tests.

djaglowski:
> posting some details from what I see with failing tests

Can you clarify exactly which test is failing? 

> these two FirstBytes are the same up until the last 2 characters

The printed bytes appear to be encoded, possibly base64. If this is the case, the string representation may be misleading, as the difference (padding) may have been introduced in the encoding process. In any case, it would be helpful to have the exact code used to create this issue.

I agree with Tigran that we really need to have a specific test that demonstrates the problem.

rockb1017:
I added more logs whenever it tails a new file.
`f.Infow("Started watching file", "path", path, "byte", reader.Fingerprint.FirstBytes, "text", string(reader.Fingerprint.FirstBytes))`
this is from running this test. 
`go test -v -run TestMultiCopyTruncateSlow`

djaglowski:
These are some pretty significant code changes, and frankly I can't recommend specifically improvements because I can't tell whether the original problem was solved or not.

If I understand correctly, you believe you have identified the root cause of the original issue. If so, I believe a reasonable path forward would be:
1. Update the original issue with a detailed explanation of what is wrong with the current implementation.
2. Start a new branch from `main`, which includes only one new test. This test should demonstrate the problem as consistently as possible. Open a draft PR to share your progress.
3. Rebase this PR onto that branch.

This approach will allow us to establish whether the changes are solving the problem.

rockb1017:
> These are some pretty significant code changes, and frankly I can't recommend specifical improvements because I can't tell whether the original problem was solved or not.
> 
> If I understand correctly, you believe you have identified the root cause of the original issue. If so, I believe a reasonable path forward would be:
> 
> 1. Update the original issue with a detailed explanation of what is wrong with the current implementation.
> 2. Start a new branch from `main`, which includes only one new test. This test should demonstrate the problem as consistently as possible. Open a draft PR to share your progress.
> 3. Rebase this PR onto that branch.
> 
> This approach will allow us to establish whether the changes are solving the problem.

Thank you for suggesting a great process. it makes sense. I have done exactly that.

operator/builtin/input/file/file.go
```diff
@@ -111,55 +118,67 @@ func (f *InputOperator) startPoller(ctx context.Context) {
...
+				}
+				if err := reader.InitializeOffset(f.startAtBeginning); err != nil {
+					f.Errorw("Failed to initialize offset for "+path, zap.Error(err))
+				}
+			} else {
+				f.Infow("Started watching file", "path", path)
+			}
+			f.tailingReaders[path] = []*Reader{reader}
+		} else { // If path exists
+			f.Infow("Log rotation detected. Started watching file", "path", path)
```
djaglowski, line 153: Does this necessarily mean the file has rotated? Wouldn't this happen any time we have seen the same file path before?

operator/builtin/input/file/config.go
```diff
@@ -151,10 +151,11 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator,
 		firstCheck:         true,
 		cancel:             func() {},
 		knownFiles:         make([]*Reader, 0, 10),
+		tailingReaders:     map[string][]*Reader{},
```
djaglowski, line 154: What does the slice of Readers represent?

djaglowski: If it is all previously rotations of the file, then do we really need to keep them? When we detect a rotation, can we just read the rest of the old file and then close it? Do we expect that rotated files are still written to? (My understanding is that the issue we have is that a file is written to immediately before being rotated.)

operator/builtin/input/file/file.go
```diff
@@ -111,55 +118,67 @@ func (f *InputOperator) startPoller(ctx context.Context) {
...
+
+	if len(f.readerQueue) <= readerCount {
+		for _, readers := range f.tailingReaders {
+			f.readerQueue = append(f.readerQueue, readers...)
+		}
+	}
+
+	count := min0(f.MaxConcurrentFiles, readerCount)
+	count = min0(count, len(f.readerQueue))
+	polledReaders := f.readerQueue[:count]
```
djaglowski, line 178: Unfortunately, I don't think `max_concurrent_files` is really being respected with this design. 

The original problem that this setting was introduced for, is when there are so many matching files that we run out of file descriptors. By leaving all the files open, we are not limiting the number of file descriptors.

I'm not certain that it's possible to accomplish both objectives though (limiting file descriptors and tracking rotations), so perhaps we would need to set a reasonably high default for `max_concurrent_files` and then only keep that many open. (This would mean that if you are matching too many files, you may lose a small percent of logs upon rotation)

djaglowski: Obviously not an ideal situation, but I think users must be able to allocate a fixed number of file descriptors to the receiver.

djaglowski (changes requested):
This PR is really complex, and there are a lot of changes that have implications I still do not fully understand. I do not see the cause of test failures. However, I left some feedback which may help the design.

operator/builtin/input/file/config.go
rockb1017: good point. once file is rotated, it won't be written to. I should add a workflow to stop tailing them yea?

operator/builtin/input/file/file.go
rockb1017: in every `poll` cycle, we will be checking the same file we have been tailing. only if the fingerprint is different and the path is what we have been tailing, then it means log rotation. 
Did i address your concern well here?

operator/builtin/input/file/file.go
rockb1017: you are right. so set it to reasonably high like like 5,000? we definitely need to have a workflow to close out already rotated files to be support this. but even then, in situation where there are 6000 actively written files and maximum is 5000, then it will only tail 5000 files and not tail the rest 1000 files at all.

PR #145, opened 2021-05-17
Add test that validates fingerprint.StartsWith when read from files

djaglowski:


operator/builtin/input/file/fingerprint_test.go
```diff
@@ -212,4 +214,49 @@ func TestFingerprintStartsWith(t *testing.T) {
 	}
 }
 
+func TestFingerprintStartsWith_FromFile(t *testing.T) {
+	rand.Seed(112358)
```
tigrannajaryan, line 218: This global may not work well if tests are run in parallel (do we?).
`r := rand.New(rand.NewSource(112358))` and using `r.Read()` should be safer.

operator/builtin/input/file/fingerprint_test.go
```diff
@@ -212,4 +214,49 @@ func TestFingerprintStartsWith(t *testing.T) {
 	}
 }
 
+func TestFingerprintStartsWith_FromFile(t *testing.T) {
```
tigrannajaryan, line 217: I am not familiar with what the fingerprints and how they work so I don't understand what this test does. Perhaps that's fine and I am not supposed to understand. If you want me to understand though please add some comments to explain what it does :-)

tigrannajaryan (approved):

operator/builtin/input/file/fingerprint_test.go
djaglowski: Good catch. Fixed.

operator/builtin/input/file/fingerprint_test.go
djaglowski: It's certainly not meant to be a secret. :) 

I expanded the comments on the `Fingerprint` struct and its `StartsWith` method, and also added a high level explanation on the new test.

tigrannajaryan (approved):
Thank you for the comments, very helpful.

PR #148, opened 2021-05-19
failing test for files rotated outside of Include path parameter

rockb1017:
#140 fixes this failing test
When tailing file get rotated to a path that it is no longer matched by `Include` path parameter, it loses trace of it. 
this is issue with kubernetes container logs because it has to tail symlinks file created by `kubelet` to parse out other metadata from file path itself.


djaglowski:
Superseded by #140

Issue #105, opened 2021-04-19
Autogenerated operator IDs


djaglowski:
Currently, operator IDs default to the operator type. If a second operator of the same type is specified, then the `id` field must be specified in order to avoid a collision.

### Motivation

When chaining together simple operations, manual assignment of `id`s becomes a cumbersome and cluttering task.

Example: 
```
operators:
  - type: move
    id: move1
    from: a
    to: b
  - type: move
    id: move2
    from: c
    to: d
  - type: move
    id: move3
    from: e
    to: f
```

### Considerations

Ideally, `id`'s could be autogenerated, but there are some considerations in pursuing this.

Most importantly, we must acknowledge that `id`'s are necessary to persist data. A violate `id` will result in the inability to recover checkpoints after a restart, or could even result in the retrieval of incorrect checkpoints (stored by another operator). For this reason, it is very important that any approach to autogenerating `id`'s is understood to be acceptably stable. Manually assigned `id`'s may be inconvenient, but they have the advantage of being entirely controllable.

Also important, is that users may encounter or even depend upon operator `id`'s, particularly when consuming signals emitted by this library.

An ideal solution would allow users to ignore the concept of `id`'s in most cases, and would minimize the risk of `id`'s changing due to minor configuration changes.

### Proposal

The simplest approach, and one that I think works reasonably well for the above considerations, is to assign `id`'s based on 1) the `type` of operator, and 2) the order in which the operators appear in the file. By example:
```
operators:
  - type: file_input  # id: file_input
    include: [ ... ]
  - type: file_input  # id: file_input1
    include: [ ... ]
  - type: move        # id: move
    from: a
    to: b
  - type: move        # id: move1
    from: c
    to: d
  - type: move        # id: move2
    from: e
    to: f
```

PR #150, opened 2021-05-19
Add Plugin ID testing

Mrod1598:
Adds testing for Plugin IDs. This is in preparation for auto-generating the IDs of the same typed operators.

### DIFF
--- a/plugin/config_test.go
+++ b/plugin/config_test.go
@@ -126,9 +126,9 @@ pipeline:
 	pluginContent2 := []byte(`
 parameters:
 pipeline:
-  - id: noop3
+  - id: noop
     type: noop
-  - id: noop4
+  - id: noop1
     type: noop
 `)
 	secondPlugin := "secondPlugin"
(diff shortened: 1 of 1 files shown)


plugin/config_test.go
```diff
@@ -101,6 +101,143 @@ output: stdout
...
+	require.NoError(t, err)
+	operator.RegisterPlugin(pluginVar.ID, pluginVar.NewBuilder)
+
+	// TODO: remove ID assignment
+	pluginContent2 := []byte(`
+parameters:
+pipeline:
+  - id: noop3
+    type: noop
+  - id: noop4
```
djaglowski, line 131: I know we're going to make these unnecessary with #105, but for now, these tests would be stronger if these id's were also `noop` and `noop1`. This would demonstrate that the scoping functionality is differentiating successfully. Also, even with autogenerated IDs, I think these will end up at `noop` and `noop1`.

djaglowski (commented):
This looks great. One suggestion to make the tests stronger.

djaglowski (approved):

Issue #151, opened 2021-05-19
Output to Plugin operator is incorrect


Mrod1598:
Currently, when trying to output to a plugin it points to the entire plugin. Which is incorrect because it doesn't exist as a singular object during processing.

Example:

Config for Plugin

```
parameters:
pipeline:
 - type: noop
    id: noop1
 - type: noop
    id: noop2
```

Pipeline
```
operators:
 - type: noop
    id: noop
 - type: plugin
    id: plugin1
```

So, when trying to output to a plugin, it runs into the issue of the previous operator, `noop`, pointing to the entire plugin with an output defined as `$.plugin1`. Which, once the operators are built and flattened, doesn't exist. What it should be pointing towards is the first operator in the plugin pipeline, in this case, it should be `$.plugin1.noop1`.

PR #156, opened 2021-05-20
Fix plugin inputs

Mrod1598:
Resolves https://github.com/open-telemetry/opentelemetry-log-collection/issues/151


pipeline/config.go
```diff
@@ -53,12 +59,27 @@ func (c Config) BuildPipeline(bc operator.BuildContext, defaultOperator operator
 	return NewDirectedPipeline(operators)
 }
 
-func getBuildContextWithDefaultOutput(configs []operator.Config, i int, bc operator.BuildContext) operator.BuildContext {
-	if i+1 >= len(configs) {
-		return bc
+func SetOutputIDs(operators []operator.Operator, pluginMap map[string]string) error {
+	for i, op := range operators {
+		if len(op.GetOutputIDs()) == 0 && i+1 < len(operators) {
```
djaglowski, line 64: For the sake of cognitive complexity, I think it would be preferable to check the `i+1 < len(operators)` condition right at the beginning of the loop, and to include a brief comment on why:
```go
// The last operator does not have a next operator to output to
if i+1 < len(operators) {
  break
}
```
Or just convert this to `for i := 0; i+1 < len(operators); i++ { ... }`

djaglowski: If I understand the rest of the function correctly, the last operator will proceed to eventually have `SetOutputIDs` called on it, which doesn't seem right. Maybe I'm misinterpreting though.

pipeline/config.go
```diff
@@ -53,12 +59,27 @@ func (c Config) BuildPipeline(bc operator.BuildContext, defaultOperator operator
...
+			}
+			continue
+		}
+		allOutputs := []string{}
+		for _, id := range op.GetOutputIDs() {
+			if pluginMap[id] != "" {
+				allOutputs = append(allOutputs, pluginMap[id])
+				continue
+			}
+			allOutputs = append(allOutputs, id)
```
djaglowski, line 77: I know I'm always asking you to flatten your code, but in this case it's probably easier not to escape early.

Also, you can capture the value from the map and check if the key exists at the same time with the `val, ok := map[key]` idiom.

```suggestion
			if pid, ok := pluginMap[id]; ok {
				id = pid
			}
			allOutputs = append(allOutputs, id)
```

pipeline/config.go
```diff
@@ -24,14 +24,20 @@ type Config []operator.Config
 // BuildOperators builds the operators from the list of configs into operators
 func (c Config) BuildOperators(bc operator.BuildContext) ([]operator.Operator, error) {
 	operators := make([]operator.Operator, 0, len(c))
-	for i, builder := range c {
-		nbc := getBuildContextWithDefaultOutput(c, i, bc)
-		op, err := builder.Build(nbc)
+	pluginMap := make(map[string]string)
```
djaglowski, line 27: Please add a comment explaining the usage of this map

operator/helper/output.go
```diff
@@ -65,10 +65,23 @@ func (o *OutputOperator) Outputs() []operator.Operator {
...
 // SetOutputs will return an error if called.
 func (o *OutputOperator) SetOutputs(operators []operator.Operator) error {
 	return errors.NewError(
 		"Operator can not output, but is attempting to set an output.",
 		"This is an unexpected internal error. Please submit a bug/issue.",
 	)
 }
+
+// SetOutputs will return an error if called.
+func (o *OutputOperator) SetOutputIDs(opIDs []string) error {
```
djaglowski, line 82: I believe this is the only circumstance where an actual error is returned, so we can probably just remove the return value from this method signature and make this a do nothing.

pipeline/config.go
```diff
@@ -53,12 +59,27 @@ func (c Config) BuildPipeline(bc operator.BuildContext, defaultOperator operator
...
+	for i, op := range operators {
+		if len(op.GetOutputIDs()) == 0 && i+1 < len(operators) {
+			err := op.SetOutputIDs([]string{operators[i+1].ID()})
+			if err != nil {
+				return err
+			}
+			continue
+		}
+		allOutputs := []string{}
+		for _, id := range op.GetOutputIDs() {
```
djaglowski, line 72: Can you clarify what this loop is doing?

pipeline/config.go
```diff
@@ -24,14 +24,20 @@ type Config []operator.Config
...
-		nbc := getBuildContextWithDefaultOutput(c, i, bc)
-		op, err := builder.Build(nbc)
+	pluginMap := make(map[string]string)
+	for _, builder := range c {
+		op, err := builder.Build(bc)
 		if err != nil {
 			return nil, err
 		}
+
+		if len(op) > 1 {
```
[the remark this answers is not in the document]
Mrod1598: the best Ideas I can think of off the top of my head other than this would be maybe trying to assert that it is Plugin and if that works then 'do work', or adding it to the entire interface which would just return false for all other ops. I was trying to avoid the latter because I've already added two new signatures to the `Operator` interface.

pipeline/config.go
Mrod1598: it checks if there are any plugins within the `outputIDs` of the operator. If there is, it changes the output to be the first op in the plugin.

pipeline/config.go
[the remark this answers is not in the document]
Mrod1598: ok added it to the builder interface so that we can determine when there are multiple outputs through the new `BuildsMultipleOps` method.

operator/builtin/transformer/router/router.go
```diff
@@ -170,6 +179,12 @@ func (p *RouterOperator) SetOutputs(operators []operator.Operator) error {
 	return nil
 }
 
+// SetOutputs will set the outputs of the router operator.
+func (p *RouterOperator) SetOutputIDs(opIDs []string) error {
+	// TODO: determine proper way to set the IDs
```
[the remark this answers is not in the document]
Mrod1598: Changed to do nothing

operator/builtin/transformer/router/router.go
```diff
@@ -157,6 +157,15 @@ func (p *RouterOperator) Outputs() []operator.Operator {
 	return outputs
 }
 
+// Outputs will return all connected operators.
```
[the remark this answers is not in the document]
Mrod1598: Updated comment

pipeline/config.go
```diff
@@ -53,12 +59,27 @@ func (c Config) BuildPipeline(bc operator.BuildContext, defaultOperator operator
 	return NewDirectedPipeline(operators)
 }
 
-func getBuildContextWithDefaultOutput(configs []operator.Config, i int, bc operator.BuildContext) operator.BuildContext {
-	if i+1 >= len(configs) {
-		return bc
+func SetOutputIDs(operators []operator.Operator, pluginMap map[string]string) error {
```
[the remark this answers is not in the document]
Mrod1598: Added comment explaining function.

operator/helper/writer.go
```diff
@@ -84,6 +81,11 @@ func (w *WriterOperator) Outputs() []operator.Operator {
 	return w.OutputOperators
 }
 
+// Outputs returns the outputs of the writer operator.
```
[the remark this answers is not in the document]
Mrod1598: Updated comment

operator/helper/output.go
```diff
@@ -65,10 +65,23 @@ func (o *OutputOperator) Outputs() []operator.Operator {
 	return []operator.Operator{}
 }
 
+// Outputs will always return an empty array for an output ID.
```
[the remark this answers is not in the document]
Mrod1598: Updated comment

operator/operator.go
```diff
@@ -40,8 +40,12 @@ type Operator interface {
 	CanOutput() bool
 	// Outputs returns the list of connected outputs.
 	Outputs() []Operator
+	// Outputs returns the list of connected outputs.
```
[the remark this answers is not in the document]
Mrod1598: Updated comment

pipeline/config.go
Mrod1598: Changed the logic around to the better logic. I'll keep this in the mind in the future thanks!

pipeline/config.go
Mrod1598: Added comment to explain map.

operator/helper/output.go
Mrod1598: Removed error from the signature.

pipeline/config.go
```diff
@@ -21,17 +21,25 @@ import (
...
-// BuildOperators builds the operators from the list of configs into operators
+// BuildOperators builds the operators from the list of configs into operators.
 func (c Config) BuildOperators(bc operator.BuildContext) ([]operator.Operator, error) {
 	operators := make([]operator.Operator, 0, len(c))
-	for i, builder := range c {
-		nbc := getBuildContextWithDefaultOutput(c, i, bc)
-		op, err := builder.Build(nbc)
+	// buildsMulti is used for storing a key of the Builder that builds multiple operators.
+	// The map is then used in SetOutputIDs to assign the output value of any ops pointing to the Builder as their output to the first operator in said Builder.
+	buildsMulti := make(map[string]string)
```
djaglowski, line 29: I find this explanation and variable name confusing. What does the key represent? What does the value represent?

djaglowski: If I understand correctly, the key is the ID of a plugin, and the value is the ID of that plugin's first operator. Is that right?

pipeline/config.go
djaglowski: Reiterating request to please add a comment explaining why the loop does not run on the last operator

Mrod1598: because no output is specified at this point for the last operator, it will always be empty and there is nothing after it to automatically point towards.

pipeline/config.go
Mrod1598: yes, I was trying to generalize it in case we add operators that build multiple ops as well.

plugin/config_test.go
```diff
@@ -238,6 +239,206 @@ pipeline:
...
+	require.NoError(t, err)
+	operator.RegisterPlugin(secondPluginVar.ID, secondPluginVar.NewBuilder)
+
+	cases := []PluginOutputIDTestCase{
+		{
+			Name: "same_op_outside_plugin",
+			PluginConfig: func() []operator.Config {
+				return pipeline.Config{
+					operator.Config{
+						// TODO: ID should be noop to start then auto gened to noop2
```
djaglowski, line 284: This is not accurate. Due to namespacing, IDs that exist within plugins should not collide with IDs at other levels or within other plugins.

This first test case should work without specifying the ID of the first operator:
```yaml
pipeline:
  - type: noop
  - type: my_plugin

# my_plugin
pipeline:
  - type: noop
    id: noop
  - type: noop
    id: noop1
    output: {{ .output }}
```

The expected pipeline would be:
`$.noop` => `$.my_plugin.noop` => `$.my_plugin.noop1`

---

Since we're also testing that the plugin ID is set correctly, let's either enhance this test or add another one that effectively does:

```yaml
pipeline:
  - type: noop
  - type: my_plugin
  - type: noop1

# my_plugin
pipeline:
  - type: noop
    id: noop
  - type: noop
    id: noop1
    output: {{ .output }}
```

The expected pipeline would be:
`$.noop` => `$.my_plugin.noop` => `$.my_plugin.noop1` => `$.noop1`

plugin/config_test.go
```diff
@@ -238,6 +239,206 @@ pipeline:
...
+									OperatorID:   secondPlugin,
+									OperatorType: secondPlugin,
+								},
+							},
+							Parameters: map[string]interface{}{},
+							Plugin:     secondPluginVar,
+						},
+					},
+					operator.Config{
+						Builder: noop.NewNoopOperatorConfig("noop5"),
```
djaglowski, line 372: Let's just have this be `noop` instead of `noop5`. It should work the same, but validates that namespacing is working correctly.

plugin/config_test.go
```diff
@@ -238,6 +239,206 @@ pipeline:
...
+									OperatorType: pluginName,
+								},
+								OutputIDs: []string{secondPlugin},
+							},
+							Parameters: map[string]interface{}{},
+							Plugin:     pluginVar,
+						},
+					},
+					operator.Config{
+						Builder: noop.NewNoopOperatorConfig("noop5"),
```
djaglowski, line 401:
```suggestion
						Builder: noop.NewNoopOperatorConfig("noop"),
```

plugin/config_test.go
```diff
@@ -238,6 +239,206 @@ pipeline:
 	}
 }
 
+type PluginOutputIDTestCase struct {
+	Name          string
+	PluginConfig  pipeline.Config
+	ExpectedOpIDs map[string][]string
+}
+
+func TestPluginOutputIDs(t *testing.T) {
```
djaglowski, line 248: Really nice tests. This is a powerful little harness, and I think we should expand on it even more because it can really lock in the functionality.

Can you think of any other test cases?

djaglowski: One that I'd like to see is having `my_plugin` feed into another instance of `my_plugin`.

For now, this will require specifying an `id` for each instance of the plugin. Once autogenerated IDs are implemented, we can validate that these IDs can be removed.

```yaml
pipeline: 
  - type: noop 
  - type: my_plugin 
  - type: my_plugin
    id: my_plugin1 # TODO remove when autogen IDs implemented 

# my_plugin 
pipeline: 
  - type: noop 
  - type: noop 
    id: noop1 # TODO remove when autogen IDs implemented 
    output: {{ .output }}
```

djaglowski: Another would be something like `operator -> plugin -> operator -> plugin`

plugin/config_test.go
Mrod1598: Enhanced the test to better show scoping works.

plugin/config_test.go
Mrod1598: Updated the name of the operator

plugin/config_test.go
Mrod1598: Updated the operator ID

plugin/config_test.go
Mrod1598: would a plugin within another plugin be applicable

Mrod1598: Also, I added that one and another one.

djaglowski: > would a plugin within another plugin be applicable

Yes, good point. Let's have some tests

plugin/config_test.go
djaglowski: It's not updated yet

plugin/config_test.go
Mrod1598: Added the additional test.

djaglowski (approved):

PR #162, opened 2021-05-24
Support byte stream data collection through 'nop' encoding

### LINKED ISSUE #138: Support byte stream data collection
See [contrib issue 3267](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/3267) for context.

This feature should apply to `file_input`, `tcp_input`, and `udp_input` operators.

Note related issues #133 and #135.

Mrod1598:
Resolves #138.

### DIFF
--- a/operator/builtin/input/file/file_test.go
+++ b/operator/builtin/input/file/file_test.go
@@ -239,6 +239,17 @@ func TestBuild(t *testing.T) {
 			require.Error,
 			nil,
 		},
+		{
+			"NopMultilineError",
+			func(f *InputConfig) {
+				f.Multiline = helper.MultilineConfig{
+					LineEndPattern: "(",
+				}
+				f.Encoding = helper.EncodingConfig{Encoding: "Nop"}
+			},
+			require.Error,
+			nil,
+		},
 	}
 
 	for _, tc := range cases {
--- a/operator/builtin/input/file/file_test.go
+++ b/operator/builtin/input/file/file_test.go
@@ -326,6 +326,21 @@ func TestReadExistingLogs(t *testing.T) {
 	waitForMessage(t, logReceived, "testlog2")
 }
 
+// TestReadUsingNopEncoding tests when nop encoding is set, that the splitfunction returns all bytes unchanged.
+func TestReadUsingNopEncoding(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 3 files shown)


operator/builtin/input/file/file_test.go
```diff
@@ -315,6 +326,21 @@ func TestReadExistingLogs(t *testing.T) {
...
 }
 
+// TestReadUsingNopEncoding tests when nop encoding is set, that the splitfunction returns all bytes unchanged.
+func TestReadUsingNopEncoding(t *testing.T) {
+	t.Parallel()
+	operator, logReceived, tempDir := newTestFileOperator(t, nil, nil)
+	operator.SplitFunc = helper.SplitNone()
+	// Create a file, then start
+	temp := openTemp(t, tempDir)
+	writeString(t, temp, "testlog1\ntestlog2\n")
```
djaglowski, line 336: Since the purpose of `encoding=nop` is to ingest `[]byte`, let's write and expect that in this test.

operator/builtin/input/file/file_test.go
```diff
@@ -315,6 +326,21 @@ func TestReadExistingLogs(t *testing.T) {
 	waitForMessage(t, logReceived, "testlog2")
 }
 
+// TestReadUsingNopEncoding tests when nop encoding is set, that the splitfunction returns all bytes unchanged.
+func TestReadUsingNopEncoding(t *testing.T) {
```
djaglowski, line 330: This test is good because it is simple. However, I think we should have another test where `max_log_size < len(contents) < 2*max_log_size`. The validation should confirm that the `len(first message) == max_log_size`, and that the second message contains the remainder of the bytes (a partial message that is less than `max_log_size`).

operator/helper/multiline.go
```diff
@@ -117,6 +121,20 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool) bufio.SplitFunc {
 	}
 }
 
+// SplitNone doesn't split any of the bytes, it reads in all of the bytes and returns it all at once. This is for when the encoding is nop
```
djaglowski, line 124: Does it respect `max_log_size`?

djaglowski (changes requested):

PR #174, opened 2021-06-09
File input should not complain about files already being closed

djaglowski:
File input generates errors when closing files, because they are being closed twice.


djaglowski:
#175 Fixes codecov config. Will rebase when it is merged.

jsirianni (approved):

Issue #179, opened 2021-06-10
Several rotation tests are failing


djaglowski:
See https://github.com/open-telemetry/opentelemetry-log-collection/pull/178/checks?check_run_id=2797160608

djaglowski:
Resolved by #182

PR #168, opened 2021-06-02
track_rotated

rockb1017:
fixes #85 
implementing https://github.com/open-telemetry/opentelemetry-log-collection/issues/85#issuecomment-852280642

in addition 
* i removed unnecessary fingerprint comparison for copytrunc cases because it is repetitive. 
* I modified file batching test case (2nd part of test) because with this new approach, the way it handles already-discovered-recently-appended log files differently.

PR #180, opened 2021-06-10
Revert "track_rotated"

djaglowski:
Reverts open-telemetry/opentelemetry-log-collection#168

See #179


jsirianni (approved):

tigrannajaryan:
Do we need to go back to the drawing board or we know what the right way to fix this is?

djaglowski:
> Do we need to go back to the drawing board or we know what the right way to fix this is?

We didn't need to go back to the drawing board, but there was some investigation required to understand the failures. 

We identified two issues and addressed them on #182. 
1. The first PR (#168) moved some code that protected against duplicated logs in Copy/Truncate scenarios. It made sense for this code to exist in the place it was moved to, but it was also needed in its original position. Thus, it is now replicated rather than moved. Performance benchmarks show trivial difference.
2. We are clarifying that the Move/Create rotation mechanism is not supported on Windows. In theory, `FILE_SHARE_DELETE` mode would allow this to work, but golang [does not support](https://github.com/golang/go/issues/32088) this. Thus, the correct action is to skip those particular tests on windows. I will document this more clearly in the design doc I am writing up.

PR #184, opened 2021-06-15
filelog: add file_name_resolved and file_path_resolved attributes

### LINKED ISSUE #181: Add real filepath to the file input
[`include_file_path`](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/operators/file_input.md#configuration-fields) includes file path, but file path is not resolved in case of symbolic links:

file_path: STRING(/xyz/test_symbolic.log)
```
lrwxrwxrwx 1 drosiek drosiek 8 Jun 11 14:41 test_symbolic.log -> test.log
```

This seems to be important metadata for us and I would like to add `include_file_real_path`

sumo-drosiek:
Add `file_name_resolved` and `file_path_resolved` attributes. They are going to keep information about absolute path and file name after symlinks resolution

fixes #181

### DIFF
--- a/operator/builtin/input/file/reader.go
+++ b/operator/builtin/input/file/reader.go
@@ -26,6 +26,7 @@ import (
 	"golang.org/x/text/encoding"
 	"golang.org/x/text/transform"
 
+	"github.com/open-telemetry/opentelemetry-log-collection/entry"
 	"github.com/open-telemetry/opentelemetry-log-collection/errors"
 )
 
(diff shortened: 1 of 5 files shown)


operator/builtin/input/file/reader.go
```diff
@@ -150,10 +151,37 @@ func (f *Reader) emit(ctx context.Context, msgBuf []byte) error {
...
 }
 
+// resolvePath resolves path to real and absolute form and add them into entry
+func (f *Reader) resolvePath(e *entry.Entry) error {
+	resolved, err := filepath.EvalSymlinks(f.Path)
+	if err != nil {
+		return err
+	}
+
+	abs, err := filepath.Abs(resolved)
```
djaglowski, line 169: Do you expect a performance hit here?

This could be computed once per Reader, but that would mean it does not detect changes in symlinks. So perhaps the next best thing would be to check if it is needed at all. Since the default configuration does not need it, this would potentially avoid a performance hit for many users.

sumo-drosiek:
Closing in favor of #184

PR #186, opened 2021-06-15
Fix default output

### LINKED ISSUE #185: Agent builder does not respect WithDefaultOutput
The agent builder has a method called `WithDefaultOutput`, which allows for specifying an output operator that will be attached to the end of a linear pipeline. This is currently not working as intended.

Expectation:
A pipeline with operators `A -> B` and default output `C`, should produce `A -> B -> C`

Result:
Currently, `B` is not connected to `C`

Mrod1598:
Resolves #185

### DIFF
--- a/agent/builder_test.go
+++ b/agent/builder_test.go
@@ -20,11 +20,35 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
+	"github.com/open-telemetry/opentelemetry-log-collection/operator"
+	"github.com/open-telemetry/opentelemetry-log-collection/operator/builtin/transformer/noop"
 	"github.com/open-telemetry/opentelemetry-log-collection/testutil"
 )
 
 func TestBuildAgentSuccess(t *testing.T) {
 	mockCfg := Config{}
+	mockLogger := zap.NewNop().Sugar()
+	mockPluginDir := "/some/path/plugins"
+
+	agent, err := NewBuilder(mockLogger).
+		WithConfig(&mockCfg).
+		WithPluginDir(mockPluginDir).
+		Build()
+	require.NoError(t, err)
+	require.Equal(t, mockLogger, agent.SugaredLogger)
+}
+
+func TestBuildAgentDefaultOperator(t *testing.T) {
+	mockCfg := Config{
+		[]operator.Config{
+			operator.Config{
+				Builder: noop.NewNoopOperatorConfig("noop"),
+			},
+			operator.Config{
+				Builder: noop.NewNoopOperatorConfig("noop1"),
+			},
+		},
+	}
 	mockLogger := zap.NewNop().Sugar()
 	mockPluginDir := "/some/path/plugins"
 	mockOutput := testutil.NewFakeOutput(t)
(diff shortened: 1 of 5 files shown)


agent/builder_test.go
```diff
@@ -36,6 +60,29 @@ func TestBuildAgentSuccess(t *testing.T) {
...
+			require.Equal(t, "$.fake", ops[1].GetOutputIDs()[0])
+			exists["$.noop1"] = true
+		} else if op.ID() == "$.fake" {
+			require.Equal(t, 0, len(ops[2].GetOutputIDs()))
+			exists["$.fake"] = true
+		}
+	}
+	require.Equal(t, true, exists["$.noop"])
+	require.Equal(t, true, exists["$.noop1"])
+	require.Equal(t, true, exists["$.fake"])
```
djaglowski, line 85: Nit: `require.True(t, exists["$.noop"])`

agent/builder_test.go
```diff
@@ -36,6 +60,29 @@ func TestBuildAgentSuccess(t *testing.T) {
...
+
+	ops := agent.pipeline.Operators()
+	require.Equal(t, 3, len(ops))
+
+	exists := make(map[string]bool)
+
+	for _, op := range ops {
+		if op.ID() == "$.noop" {
+			require.Equal(t, 1, len(ops[0].GetOutputIDs()))
+			require.Equal(t, "$.noop1", ops[0].GetOutputIDs()[0])
```
djaglowski, line 72:
```suggestion
			require.Equal(t, 1, len(op.GetOutputIDs()))
			require.Equal(t, "$.noop1", op.GetOutputIDs()[0])
```

agent/builder_test.go
```diff
@@ -36,6 +60,29 @@ func TestBuildAgentSuccess(t *testing.T) {
...
+			exists["$.noop"] = true
+		} else if op.ID() == "$.noop1" {
+			require.Equal(t, 1, len(ops[1].GetOutputIDs()))
+			require.Equal(t, "$.fake", ops[1].GetOutputIDs()[0])
```
djaglowski, line 76:
```suggestion
			require.Equal(t, 1, len(op.GetOutputIDs()))
			require.Equal(t, "$.fake", op.GetOutputIDs()[0])
```

agent/builder_test.go
```diff
@@ -36,6 +60,29 @@ func TestBuildAgentSuccess(t *testing.T) {
...
+			exists["$.noop1"] = true
+		} else if op.ID() == "$.fake" {
+			require.Equal(t, 0, len(ops[2].GetOutputIDs()))
```
djaglowski, line 79:
```suggestion
			require.Equal(t, 0, len(op.GetOutputIDs()))
```

pipeline/config.go
```diff
@@ -38,6 +38,12 @@ func (c Config) BuildOperators(bc operator.BuildContext) ([]operator.Operator, e
 		}
 		operators = append(operators, op...)
 	}
+
+	if defaultOperator != nil {
+		bc.DefaultOutputIDs = []string{defaultOperator.ID()}
```
djaglowski, line 43: BuildContext is not used after this point, so this line can be removed.

pipeline/config.go
```diff
@@ -38,6 +38,12 @@ func (c Config) BuildOperators(bc operator.BuildContext) ([]operator.Operator, e
 		}
 		operators = append(operators, op...)
 	}
+
+	if defaultOperator != nil {
```
djaglowski, line 42: We should also check whether the last operator can output. Otherwise, appending the default operator will have no effect other than inserting an orphaned node into the graph.

```suggestion
	if defaultOperator != nil && operators[len(operators)-1].CanOutput() {
```

djaglowski (changes requested):

pipeline/config.go
Mrod1598: fixed

pipeline/config.go
Mrod1598: Added

agent/builder_test.go
Mrod1598: Changed

agent/builder_test.go
Mrod1598: Changed

agent/builder_test.go
Mrod1598: Changed

agent/builder_test.go
Mrod1598: Changed

rockb1017:
awesome. great the root cause is already figured out! thank you!

djaglowski (approved):
LGTM. Thanks @Mrod1598!

PR #189, opened 2021-06-16
filelog: add file_name_resolved and file_path_resolved attributes

### LINKED ISSUE #181: Add real filepath to the file input

sumo-drosiek:
Add `file_name_resolved` and `file_path_resolved` attributes. They are going to keep information about absolute path and file name after symlinks resolution

fixes #181 
Replacement for #184 (new head repository)

### DIFF
--- a/operator/builtin/input/file/config.go
+++ b/operator/builtin/input/file/config.go
@@ -37,15 +37,17 @@ const (
 // NewInputConfig creates a new input config with default values
 func NewInputConfig(operatorID string) *InputConfig {
 	return &InputConfig{
-		InputConfig:        helper.NewInputConfig(operatorID, "file_input"),
-		PollInterval:       helper.Duration{Duration: 200 * time.Millisecond},
-		IncludeFileName:    true,
-		IncludeFilePath:    false,
-		StartAt:            "end",
-		FingerprintSize:    defaultFingerprintSize,
-		MaxLogSize:         defaultMaxLogSize,
-		MaxConcurrentFiles: defaultMaxConcurrentFiles,
-		Encoding:           helper.NewEncodingConfig(),
+		InputConfig:             helper.NewInputConfig(operatorID, "file_input"),
+		PollInterval:            helper.Duration{Duration: 200 * time.Millisecond},
+		IncludeFileName:         true,
+		IncludeFilePath:         false,
+		IncludeFileNameResolved: false,
+		IncludeFilePathResolved: false,
+		StartAt:                 "end",
+		FingerprintSize:         defaultFingerprintSize,
+		MaxLogSize:              defaultMaxLogSize,
+		MaxConcurrentFiles:      defaultMaxConcurrentFiles,
+		Encoding:                helper.NewEncodingConfig(),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


sumo-drosiek:
[Original comment](https://github.com/open-telemetry/opentelemetry-log-collection/pull/184#discussion_r651777908)
> Do you expect a performance hit here?
> 
> This could be computed once per Reader, but that would mean it does not detect changes in symlinks. So perhaps the next best thing would be to check if it is needed at all. Since the default configuration does not need it, this would potentially avoid a performance hit for many users.

@djaglowski

I moved path resolving to Reader init. I also added test for rotation case and it looks fine. There is no issue with rotation due to implemented fingerprint.

One thing I observed during test, is that if we rotate to file with the same content it won't work. Is that a serious issue? Should it be fixed?

operator/builtin/input/file/reader.go
```diff
@@ -29,15 +29,44 @@ import (
...
+func resolveFileAttributes(path string) *fileAttributes {
+	resolved, err := filepath.EvalSymlinks(path)
+	if err != nil {
+		resolved = ""
+	}
+
+	abs, err := filepath.Abs(resolved)
+	if err != nil {
+		abs = ""
+	}
```
djaglowski, line 51: Best I can tell, both `filepath.EvalSymlinks` and `filepath.Abs` will only ever return `"", err`, or `"not empty", nil`, so these error checks aren't really doing anything.

Moreover, I think we should return the errors and log them in the callers. In both cases (`newReader` and `NewReader`), we have a handle to the operator, which has a logger.

operator/builtin/input/file/file_test.go
```diff
@@ -65,6 +66,128 @@ func TestAddFileFields(t *testing.T) {
...
+	// Ignore error on windows host (The process cannot access the file because it is being used by another process.)
+	os.RemoveAll(dir)
+}
+
+// AddFileResolvedFields tests that the `file_name_resolved` and `file_path_resolved` fields are included
+// when IncludeFileNameResolved and IncludeFilePathResolved are set to true and underlaying symlink change
+// Scenario:
+// monitored file (symlink) -> middleSymlink -> file_1
+// monitored file (symlink) -> middleSymlink -> file_2
+func TestAddFileResolvedFieldsWithChangeOfSymlinkTarget(t *testing.T) {
```
djaglowski, line 120: Very nice test. 

Am I correct in understanding that the reason this works, is because each poll interval creates a new Reader, which re-resolves the symlink?

djaglowski (commented):
Overall, this looks very nice. A few suggestions though.

Additionally, I think we should look at performance impact. Will you please run `make bench` on `main` and on your branch, and post the relevant benchmark results here?

> if we rotate to file with the same content it won't work

This is as designed. Files with exactly the same contents are typically not meant to be reingested. Most commonly, this occurs when files are rotated using the copy/truncate method. In this scenario, multiple files with the same fingerprint exist briefly (after copy, before truncate), but they are deduplicated so as not to duplicate ingestion. One could perhaps define a use case that requires reingestion of the same content under a different file name, but I'm somewhat skeptical that this would be of enough value to support automatically.

operator/builtin/input/file/file_test.go
sumo-drosiek: If path or fingerprint changes, reader is being cloned with new path

operator/builtin/input/file/reader.go
sumo-drosiek: Not sure how to properly implement it. For now all errors will result in no data ingestion:

https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/builtin/input/file/file.go#L283-L286

Should I exclude filepath related errors from this condition and just print logs for them?

djaglowski: I see your point. Probably then it makes sense to just make this a receiver on the operator:

```go
func (f *InputOperator) resolveFileAttributes(path string) *fileAttributes {
	resolved, err := filepath.EvalSymlinks(path)
	if err != nil {
		f.Error(err)
	}

	abs, err := filepath.Abs(resolved)
	if err != nil {
		f.Error(err)
	}

	return ...
}
```

operator/builtin/input/file/file_test.go
```diff
@@ -65,6 +66,128 @@ func TestAddFileFields(t *testing.T) {
...
+	writeString(t, file2, "testlog2\n")
+
+	e = waitForOne(t, logReceived)
+	require.Equal(t, filepath.Base(symLinkPath), e.Attributes["file_name"])
+	require.Equal(t, symLinkPath, e.Attributes["file_path"])
+	require.Equal(t, filepath.Base(resolved2), e.Attributes["file_name_resolved"])
+	require.Equal(t, resolved2, e.Attributes["file_path_resolved"])
+
+	// Clean up (linux based host)
+	// Ignore error on windows host (The process cannot access the file because it is being used by another process.)
```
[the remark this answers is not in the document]
sumo-drosiek: Using operator.Stop() before calling this didn't worked :disappointed:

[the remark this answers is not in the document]
djaglowski: Ok, thanks for trying. Doesn't need to hold this up

operator/builtin/input/file/config.go
```diff
@@ -138,24 +142,36 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator,
 		filePathField = entry.NewAttributeField("file_path")
 	}
 
+	fileNameResolvedField := entry.NewNilField()
+	if c.IncludeFileNameResolved {
+		fileNameResolvedField = entry.NewAttributeField("file_name_resolved")
```
tigrannajaryan, line 147: I think we should look into standardizing file related attribute names as Otel semantic conventions.

sumo-drosiek: valid point. Are they any ideas for that? Should it be added to https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions?

tigrannajaryan: Yes. I think we need to decide on how we group this attributes, perhaps introduce a `file.* namespace and put everything there. Perhaps also some of the attributes discussed [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2628) should in that namespace, e.g. `stream`?

djaglowski: I suggest we do the following:
- In this PR, choose attributes names we think are likely to be appropriate. Update this PR accordingly. Expect that they may have to change.
- File a PR against the spec repo, which proposes these names as semantic conventions.

Does this seem like a reasonable approach? 

--- 

Am I correct in thinking that there aren't currently any semantic conventions for log attributes, so this would also entail establishing a new section in the spec repo? Possibly parallel to [here](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions)?

djaglowski: > Should it be added to https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions?

Maybe it's just a question of where is the right place to add the conventions, but this particular location seems to imply that we are defining a file as a resource. I'm not opposed to this, but my understanding is that, in the context of the `file_input` operator / `filelog` receiver, we should not consider files to be resources. They are not the things that emitted the logs. They are essentially just a mechanism by which logs are transmitted.

So just to be clear, are we talking about defining file-related fields as a resource, but then just using the same convention to structure our attributes here? Or do we need to add a parallel section to the spec, which specifically defines attribute conventions for files?

Maybe this isn't an important nuance, but I want to make sure we're not missing it.

djaglowski: First pass proposal for a new `file.*` namespace:
```
file.name
file.path
file.name.resolved
file.path.resolved
file.stream
```

djaglowski (approved):
This looks good to me.

We have one outstanding item (tracked in #191) to address regarding the establishment of a `file.*` namespace in the semantic conventions, but I believe we are using reasonable attribute names and can adapt them later if necessary. 

Any objections to merging now @tigrannajaryan?

operator/builtin/input/file/config.go
tigrannajaryan: I think you are right, these are not resource attributes, they are log record attributes.

tigrannajaryan:
> This looks good to me.
> 
> We have one outstanding item (tracked in #191) to address regarding the establishment of a `file.*` namespace in the semantic conventions, but I believe we are using reasonable attribute names and can adapt them later if necessary.
> 
> Any objections to merging now @tigrannajaryan?

No objection. Please file a spec repo issue and add to the next Log SIG agenda.

djaglowski:
Opened [opentelemetry-specification#1772](https://github.com/open-telemetry/opentelemetry-specification/issues/1772)

PR #201, opened 2021-06-22
Fix for Syslog output & Add testing for it

### LINKED ISSUE #200: Syslog_Input operator incorrectly connecting to the next operator.
This comes from how default operators now connect. 
The Syslog input is connecting to itself because it builds multiple operators and outputs them out of order. 
So, it outputs them as 
```
1 . ) syslog_parser
2 . ) udp_input || tcp_input
```
What it should output as is
```
1 . ) udp_input || tcp_input
2 . ) syslog_parser
```
This is because when the default outputs are set, they are done sequentially so it points to the incorrect operators to output too.

Mrod1598:
Resolves #200.


djaglowski (approved):
LGTM.

Confirmed that this fixes the issue observed when integrating into contrib.

PR #203, opened 2021-06-23
Add design documentation for file_input operator

djaglowski:


operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
+# Fingerprints
+
+Files are identified and tracked using fingerprints. A fingerprint is the first `N` bytes of the file, with the default for `N` being `1000`. 
+
+### Fingerprint Growth
+
+When a file is smaller than `N` bytes, the fingerprint is the length of the file. A fingerprint that is less than `N` bytes will be compared to other fingerprints using a prefix check. As the file grows, its fingerprint will be updated, until it reaches the full size of `N`.
```
tigrannajaryan, line 7: >the fingerprint is the length of the file
Is fingerprint literally the length? Or the fingerprint is the entire file when the file length is < N?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+The file system is polled on a regular interval, defined by the `poll_interval` setting. 
+
+Each poll cycle runs through a series of steps which are presented below.
+
+At a very high level, each poll cycle operates as three phases:
+1. Finish work that was started in the previous poll cycle.
+2. Begin work that will carry over to the next cycle.
+3. Allow some time to pass.
+
+Because state is carried over from one poll cycle to the next, the following detailed description is presented starting mid-cycle, at a point where there is no preexisting state.
```
tigrannajaryan, line 67: This sentence is not quite clear. Can you rephrase?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+        - The current poll interval will begin processing the first set of files, just as if they were the only ones found during the matching phase.
+        - Subsequent poll cycles will pull matches off of the queue, until the queue is empty.
+        - The `max_concurrent_files` setting is respected at all times.
+        - TODO: Use a worker pool to process the queue as quickly as possible, rather than waiting for poll cycles to trigger batching.
+3. Opening
+    1. Each of the matched files is opened. Note:
+        - A small amount of time has passed since the file was matched.
+        - It is possible that is has been moved or deleted by this point.
+        - Only a minimum set of operations should occur between file matching and opening.
+        - If an error occurs while opening, it is logged.
```
tigrannajaryan, line 91: Does this mean every time a log rotation via renaming happens we are likely to log an error message?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+    1. Empty files are closed immediately and discarded. (There is nothing to read.)
+    2. Fingerprints found in this batch are cross referenced against each other to detect duplicates. Duplicate files are closed immediately and discarded.
+        - In the vast majority of cases, this occurs during file rotation that uses the copy/truncate method. (See fingerprinting section above.)
+6. Reader Creation
+    1. Each file handle is wrapped into a `Reader` along with some metadata. (See Reader section above)
+        - During the creation of a `Reader`, the file's fingerprint is crossreferenced with previously known fingerprints.
+        - If a file's fingerprint matches one that has recently been seen, then metadata is copied over from the previous iteration of the Reader. Most importantly, the offset is accurately maintained in this way.
+        - If a file's fingerprint does not match any recently seen files, then its offset is initialized according to the `start_at` setting.
+7. Detection of "Lost" Files
+8. Consumption
```
tigrannajaryan, line 104: Should this be 7-8 or there is a numbering issue?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+    1. Lost files are consumed. Since these files are thought to have been rotated to a location where we will not match them again, we will never see them again.
```
tigrannajaryan, line 105: What are "lost" files?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+        - In the vast majority of cases, this occurs during file rotation that uses the copy/truncate method. (See fingerprinting section above.)
+6. Reader Creation
+    1. Each file handle is wrapped into a `Reader` along with some metadata. (See Reader section above)
+        - During the creation of a `Reader`, the file's fingerprint is crossreferenced with previously known fingerprints.
+        - If a file's fingerprint matches one that has recently been seen, then metadata is copied over from the previous iteration of the Reader. Most importantly, the offset is accurately maintained in this way.
+        - If a file's fingerprint does not match any recently seen files, then its offset is initialized according to the `start_at` setting.
+7. Detection of "Lost" Files
+8. Consumption
+    1. Lost files are consumed. Since these files are thought to have been rotated to a location where we will not match them again, we will never see them again. 
+        - The best we can do is finish consuming their current contents.
```
tigrannajaryan, line 106: Unclear how we consume them if they are lost. Is it assumed that they are still open in a Reader but will be lost the moment we close them?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+        - We can reasonable expect in most cases that these files are no longer being written to.
```
tigrannajaryan, line 107:
```suggestion
        - We can reasonably expect in most cases that these files are no longer being written to.
```

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+    2. Matched files (from this poll cycle) are consumed.
+        - These file handles will be left open until the end of the next poll cycle.
```
tigrannajaryan, line 109: end of next poll cycle or end of this poll cycle? Do they stay open for the duration of the interval between cycles plus 2*duration of reading in each cycle?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+        - Typically, we can expect to find most of these files again. However, these files are consumed greedily, in case we do not see them again.
+    3. All files in both sets are consumed concurrently.
```
tigrannajaryan, line 111: Both sets = "lost" and "matched"?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+1. Finish work that was started in the previous poll cycle.
+2. Begin work that will carry over to the next cycle.
+3. Allow some time to pass.
+
+Because state is carried over from one poll cycle to the next, the following detailed description is presented starting mid-cycle, at a point where there is no preexisting state.
+
+
+### Detailed Poll Cycle
+
+_Note: The following steps are presented starting mid poll cycle._
```
tigrannajaryan, line 72: I am a bit confused: why isn't the description bellow start from "Start Poll Cycle" step? I understand it is cyclical but it seems a bit more natural than starting mid cycle?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+    1. All files from the previous poll cycle are closed.
+10. Archiving
+    1. Readers created in the current poll cycle are added to the historical record.
+11. Pruning
+    1. The historical record is purged of Readers that have existsed for 3 generations.
+        - This number is somewhat arbitrary, and should probably be made configurable. However, its exact purpose is quite obscure.
+12. Persistence
+    1. The historical record of readers is synced to whatever persistence mechanism was provided to the operator.
+13. End Poll Cycle
+    1. At this point, the operator sits idle until the poll timer fires again.
```
tigrannajaryan, line 122: Would it make more sense to eliminate the idling if the queue is not empty so that we do the remaining work asap?

operator/builtin/input/file/design.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Startup Logic
+
+Whenever the operator starts, it:
+- Requests the historical record of Readers, as described in steps 10-12 of the poll cycle.
+- Starts the polling timer.
+
+### Shutdown Logic
+
+When the operator shuts down, it:
```
tigrannajaryan, line 140: Do we shutdown midcycle or always wait for the cycle to complete? Is there a chance to store the checkpoint before the data is actually consumed and shutdown happens?

tigrannajaryan (commented):
@djaglowski This is great, thanks for adding! Future generations of contributors will be grateful.

operator/builtin/input/file/design.md
djaglowski: Good point. Will rephrase

operator/builtin/input/file/design.md
djaglowski: I need to elaborate on the concept of lost files, and how we detect them.

operator/builtin/input/file/design.md
djaglowski: I will elaborate

operator/builtin/input/file/design.md
djaglowski: Yes, will rephrase

operator/builtin/input/file/design.md
djaglowski: Point taken. I'll reorganize it to just describe the cycle from start to finish.

I thought it might be more confusing to start the description with steps that are dependent on the state of the previous cycle, since understanding these steps requires understanding the rest of the poll cycle. 

Later, I'll explore whether these first few steps can just be moved to the end of the cycle, which would be a lot easier to reason about anyways.

operator/builtin/input/file/design.md
djaglowski: Yeah, I actually realized the same when drafting this and had added a TODO in the `Queueing` section. Basically, I think we could use a worker pool to consume the queue.

operator/builtin/input/file/design.md
djaglowski: I'll update the doc, but to answer your question: 

If a poll cycle is _not_ currently underway, the operator simply closes any open files. 

Otherwise, the current poll cycle is signaled to stop immediately, which in turn signals all Readers to stop immediately.

If a Reader is idle or in between log entries, it will return immediately. Otherwise is will return after consuming one final log entry.

When all Readers return, the remainder of the poll cycle completes as usual, which includes the steps labeled Closing, Archiving, Pruning, and Persistence.

The net effect of all this is that all files are checkpointed in a normal manner (i.e. not in the middle of a log entry), and all checkpoints are persisted.

operator/builtin/input/file/design.md
djaglowski: I'll remove this entirely, as part of reordering the steps in the cycle.

operator/builtin/input/file/design.md
djaglowski: Some other rare errors are likely, possible which we would not mind logging.

Those aside, I believe we would only expect to see this if the following is true:
1. Running on Windows
2. Logs going to a file rather than Windows Event Log
3. Logs are being rotated using move/create strategy

I'm sure all of this happens, but in theory the log emitter should make different choices. Of course this isn't always in the control of the user, so perhaps this would be an annoying log in that case.

Do you agree with the above assessment? If so, do you think logging an error is reasonable?

operator/builtin/input/file/design.md
djaglowski: Yes, that's the idea. 

By keeping the file open until the next poll cycle, we can compare two sets of "matched" files. Those that were found in the previous cycle but not in the current are considered lost. Several things may have happened, but in each case, the best we can do is try to read the file one last time.

If the file was deleted, then reading will be a nop. Typically, this would be due to rotation limits or ttl-based pruning, which means the older files have probably not been written to in a while, so we've fully consumed them.

If the file was moved, as is the case on k8s, then we can still read the remainder of it before closing. It's unlikely to be written to after being moved, so we'll have fully consumed it.

If the file was copied and truncated, we could potentially miss some lines. Rotation would have to follow essentially the same conditions as on k8s, but using copy/truncate. I'm not sure how likely this is to ever happen, but it does seem theoretically possible.

operator/builtin/input/file/design.md
djaglowski: Hmm, I need to tweak this a bit. 

This is the longest a file may remain open, but it would only occur when a file is detected as "lost".

In such a case, the lifecycle of the file handle would be:

Poll Cycle N:
- File X is matched, opened, consumed.

Poll Cycle N+1:
- File X is not matched. 
- Cross referencing with files from Poll Cycle N shows that File X has been "lost".
- Remainder of File X is consumed.
- File X is closed.

The typical file handle lifecycle would be a bit shorter:

Poll Cycle N:
- File X is matched, opened, consumed.

Poll Cycle N+1:
- File X is matched again.
- Cross referencing with files from Poll Cycle N shows we now have duplicate handles to File X.
- Close old handle to File X.

operator/builtin/input/file/design.md
tigrannajaryan: > Do you agree with the above assessment? If so, do you think logging an error is reasonable?

I think it is reasonable. We can rethink this in the future if we find that the users get scared from the error messages.

tigrannajaryan (approved):

PR #212, opened 2021-07-06
feat: trim whitechars for file with multiline

sumo-drosiek:
Strip whitechars if using multiline

### DIFF
--- a/operator/helper/multiline.go
+++ b/operator/helper/multiline.go
@@ -79,7 +79,9 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool) bufio.SplitFunc {
 		if firstLoc == nil {
 			// Flush if no more data is expected
 			if len(data) != 0 && atEOF && flushAtEOF {
-				return len(data), data, nil
+				token = trimWhitespaces(data)
+				advance = len(token)
+				return
 			}
 			return 0, nil, nil // read more data and try again.
 		}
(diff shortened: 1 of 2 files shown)


operator/helper/multiline.go
```diff
@@ -79,7 +79,9 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool) bufio.SplitFunc {
 		if firstLoc == nil {
 			// Flush if no more data is expected
 			if len(data) != 0 && atEOF && flushAtEOF {
-				return len(data), data, nil
+				token = trimWhitespaces(data)
+				advance = len(token)
```
djaglowski, line 83: `advance` is used to set the starting point for the next read. By advancing after trimming, I think we may be setting up the value such that we will reread the whitespace, which doesn't seem ideal. Unless I'm misunderstanding some part of this, then I would expect that we do:
```suggestion
				token = trimWhitespaces(data)
				advance = len(data)
```

operator/helper/multiline.go
```diff
@@ -187,3 +195,10 @@ func encodedCarriageReturn(encoding encoding.Encoding) ([]byte, error) {
 	nDst, _, err := encoding.NewEncoder().Transform(out, []byte{'\r'}, true)
 	return out[:nDst], err
 }
+
+func trimWhitespaces(data []byte) []byte {
+	// TrimLeft to strip EOF whitespaces in case of using $ in regex
+	// For some reason newline and carriage return are being moved to beginning of next log
+	// TrimRight to strip all whitespaces from the end of log
```
djaglowski, line 202: Do the test cases cover all of these behaviors? If not, should we have a few more? It seems like these nuances would be easy to miss if someone makes a future change.

sumo-drosiek: Yes, if you look at the test changes, both of the cases are tested. Should I write dedicated tests for them, so it would be clear that we are explicitly testing it?

djaglowski: I misunderstood when I looked earlier. I think the current structure is good.

Sorry, but one more ask - Can you add a test or two where there are multiple returns in between the start/end pattern?

djaglowski (approved):

operator/helper/multiline.go
sumo-drosiek: Sure, please see new commit

Issue #206, opened 2021-06-24
windows_eventlog_input: Nicer error when metadata request fails


jsirianni:
Windows event log input has two stages
1. Read events from Windows Event
2. Request metadata from Windows Events pertaining to the event in step 1.

Step 2 can fail if the event is old enough, this will produce an error like this:
```
Failed to open publisher: failed to open publisher handle: The system cannot find the file specified."
```

This error indicates a failure and does not mention that the event will still be processed in the pipeline, just without metadata. It would be nice if the error included information like "writing log entry to pipeline without metadata". 

```go
	publisher := NewPublisher()
	if err := publisher.Open(simpleEvent.Provider.Name); err != nil {
                // error is produced here
		e.Errorf("Failed to open publisher: %s")
                // event is written to the pipeline here
		e.sendEvent(ctx, simpleEvent)
		return
	}
	defer publisher.Close()
```

See https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/builtin/input/windows/operator.go#L204

This error is easily re-produced on windows systems with old events using the `start_at: beginning` parameter.

djaglowski:
Resolved by #206

PR #214, opened 2021-07-08
Add nicer error message when metadata request fails (#206)

lastchiliarch:
Add nicer error message when metadata request fails (#206)

Signed-off-by: lastchiliarch <lastchiliarch@163.com>


jsirianni (approved):
Thank you 👍

PR #216, opened 2021-07-09
Add possibility to force flush logs after certain period of time

### LINKED ISSUE #215: file_input: last log is not being flushed if multiline is used
If using [multiline detection](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/operators/file_input.md#multiline-configuration) with `line_start_pattern` for log input, last line is not being flushed at all. This is due to no control mechanism which check if more data is going to be appended to the file, so we always wait for more data.

Same behavior for no multiline and file ended with no new line

My proposal is to flush data if there was no more data appended to a file and/or the file has not been modified for some period of time. Another approach is to flush the last log if there is no file descriptors in system, which uses that file (except our). Time could be configurable in `file_input`.

---

test.log:
```
LOGSTART 1
LOGPART 1
LOGPART 1
LOGSTART 2
LOGPART 2
LOGSTART 3
```

config.yaml
```
receivers:
  filelog:
    start_at: beginning
    include:
      - ./open-receiver/test.log
    multiline:
      line_start_pattern: ^LOGSTART

exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    logs:
      receivers:
        - filelog
      exporters:
        - logging

```

logging output:
```
2021-07-09T10:05:09.084+0200    DEBUG   loggingexporter/logging_exporter.go:141 ResourceLog #0
InstrumentationLibraryLogs #0
InstrumentationLibrary  
LogRecord #0
Timestamp: 2021-07-09 08:05:08.98527233 +0000 UTC
Severity: Undefined
ShortName: 
Body: LOGSTART 2
LOGPART 2
Attributes:
     -> file.name: STRING(test.log)
LogRecord #1
Timestamp: 2021-07-09 08:05:08.985249861 +0000 UTC
Severity: Undefined
ShortName: 
Body: LOGSTART 1
LOGPART 1
LOGPART 1
Attributes:
     -> file.name: STRING(test.log)

^C2021-07-09T10:05:25.293+0200  info    service/application.go:210      Received signal from OS {"signal": "interrupt"}
```

sumo-drosiek:
Fixes #215 

This is work in progress PR. General idea is to keep information about when last log has been properly flushed and if it was long ago, buffered line should be flushed as we do not except more lines

I'm going to make minor refactor, add unit tests and more comments to code


operator/helper/multiline.go
```diff
@@ -19,62 +19,125 @@ import (
...
+}
+
+type Multiline struct {
+	SplitFunc   bufio.SplitFunc
+	force       *ForceFlush
+	forcePeriod time.Duration
+
+	// lastFlush > force.LastFlush => we can force flush if no logs are incoming for forcePeriod
+	// lastFlush = force.LastFlush => last flush was forced, so we do cannot force, we can update lastFlush
+	// lastFlush < force.LastFlush =>we just forced flush, set lastFlush to force.LastFlush
```
pmalek-sumo, line 46:
```suggestion
	// lastFlush = force.LastFlush => last flush was forced, so we cannot force, we can update lastFlush
	// lastFlush < force.LastFlush => we just forced flush, set lastFlush to force.LastFlush
```

operator/helper/multiline.go
```diff
@@ -19,62 +19,125 @@ import (
...
+		m.force.Force = true
+	}
+}
+
 // NewBasicConfig creates a new Multiline config
 func NewMultilineConfig() MultilineConfig {
 	return MultilineConfig{
 		LineStartPattern: "",
 		LineEndPattern:   "",
+		ForceFlushPeriod: "0s",
```
pmalek-sumo, line 71: What does force flush period of `0s` mean? Maybe we document that?

operator/helper/multiline.go
```diff
@@ -19,62 +19,125 @@ import (
...
+// Flushed update lastFlush with current timestamp
+func (m *Multiline) Flushed() {
+	if m.lastFlush.Sub(m.force.LastFlush) < 0 {
+		m.lastFlush = m.force.LastFlush
+	} else {
+		m.lastFlush = time.Now()
+	}
+}
+
+// CheckAndFlush returns true if data is going to be force flushed
```
pmalek-sumo, line 59: It doesn't actually return anything. It sets the flag.

sumo-drosiek: fixed

operator/helper/multiline.go
sumo-drosiek: fixed

operator/builtin/input/file/config.go
```diff
@@ -156,7 +157,7 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator,
 		InputOperator:         inputOperator,
 		Include:               c.Include,
 		Exclude:               c.Exclude,
-		SplitFunc:             splitFunc,
+		Multiline:             c.Multiline,
```
djaglowski, line 160: Perhaps this is just a naming issue, but I'll have a difficult time accepting that the splitfunc is being delegated to a `Multiline` struct. Put another way, we need a split func regardless of whether or not we are using multiline splitting. 

I suspect that we can accomplish the same functionality as you are implementing, but by composing things in a better way.

Would it make sense to create a new struct which would pair the concept of a split func with the concept of a flush? Maybe something like:
```go
type Splitter struct {
    splitFunc // can be simple like '\n' or multiline config
    flusher     // forces a flush based on timing
}
```

docs/operators/file_input.md
```diff
@@ -37,7 +37,12 @@ If set, the `multiline` configuration block instructs the `file_input` operator
 The `multiline` configuration block must contain exactly one of `line_start_pattern` or `line_end_pattern`. These are regex patterns that
 match either the beginning of a new log entry, or the end of a log entry.
 
-Also refer to [recombine](/docs/operators/recombine.md) operator for merging events with greater control. 
+If using multiline, last log can sometimes be not flushed due to waiting for more content.
+In order to forcefully flush last buffered log after certain period of time,
+set `force_flush_period` option to [duration string](https://golang.org/pkg/time/#ParseDuration),
+eg: `5s`, `1m`. It's by default `0s` which means, that no force flushing will be performed.
```
djaglowski, line 43: For consistency, I think we should use `helper.Duration` instead of `time.Duration`. We can then link to `https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/types/duration.md`, which explains formatting sufficiently.

This is basically the same thing, but with a default unit of seconds, since nanoseconds is basically never reasonable in this context.

operator/builtin/input/file/reader.go
```diff
@@ -116,7 +120,7 @@ func (r *Reader) ReadToEnd(ctx context.Context) {
 		return
 	}
 
-	scanner := NewPositionalScanner(r, r.fileInput.MaxLogSize, r.Offset, r.fileInput.SplitFunc)
+	scanner := NewPositionalScanner(r, r.fileInput.MaxLogSize, r.Offset, r.multiline.SplitFunc)
```
djaglowski, line 123: I believe this is the only place we are using the split func, so I'm not sure why it needs to be part of the multiline config.

We have two concepts here:
1. How to split lines
2. When to force a flush

I'm not convinced these are dependent on each other, or that either is necessarily a concern of multiline behavior. (Of course a split func may implement multiline splitting, but it does not always.)

Can we leave splitFunc as it was and just add a new config that controls flushing?

operator/builtin/input/file/config.go
djaglowski: After looking at this further, I'm not convinced we even need to do what I've suggested here. Please see my related comment on `reader.go`. I'll leave this comment in place as a possible alternative, but I think we need to start by justifying the need to alter splitFunc behavior in any way.

operator/builtin/input/file/reader.go
sumo-drosiek: There is some relations I was unsure how to resolve, so I put both functionality into one struct:

 - splitFunc needs to be aware when to force a flush (so new structure is created and pass to the splitFunc)
 - every file should be forced separately (so splitFunc is created for every unique reader)
 
 Regarding line splitting. Of course it is not always a multiline behavior, but looking at it from user perspective it is intuitive to use `multline` config for it, so `Multiline` struct underneath, than mixing `Splitter` struct with `multiline` config, or using `splitter` config.
 
 I will exclude force a flush to separate config, as logic of it is mostly out of the splitFunc, but not sure how to resolve other dependencies in nice way

sumo-drosiek: I made some changes in requested direction, but after all I think `Splitter` struct seems to be reasonable consensus. 

You can see current implementation without `Splitter`. There are two dependent entities: `splitFunc` and `ForceFlush` and I feel that they should be connected together via additional struct. WDYT?

operator/builtin/input/file/config.go
```diff
@@ -117,7 +119,8 @@ func (c InputConfig) Build(context operator.BuildContext) ([]operator.Operator,
 		return nil, err
 	}
 
-	splitFunc, err := c.Multiline.Build(context, encoding.Encoding, false)
+	// Ensure that multiline is buildable
+	_, err = c.Multiline.Build(encoding.Encoding, false, nil)
```
djaglowski, line 123: I'm trying to figure out why `flushAtEOF` wouldn't solve this problem. Likely I am missing something, but it seems like this should take care of it.

Of course, we don't want this enabled by default, but if it would work, then we should be able to solve this by just adding a boolean parameter and passing it through here.

Thoughts on this?

sumo-drosiek: In case log is written in multiple parts, `flushAtEOF` will result in flushing every part separately. [link to test](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/builtin/input/file/file_test.go#L327-L344) I think we are replacing one edge case with another one

If I set `flushAtEOF` to `true` for file input, file is re-read every time due to some fingerprinting issue, which probably needs to be investigated.

sumo-drosiek: I tested it with `fingerprint_size: 20` and what i observed for this config
```
receivers:
  filelog:
    start_at: beginning
    include:
      - ./open-receiver/test.log
    multiline:
      line_start_pattern: 'START '
      # force_flush_period: 3s
    fingerprint_size: 20

exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    logs:
      receivers:
        - filelog
      exporters:
        - logging
```

adding log to the file (copying and pasting `asd\n`) results sometimes with `asd` log, sometimes with  `asd\nasd`

djaglowski: Thanks for clarifying.

operator/builtin/input/file/reader.go
djaglowski: I think consolidating into one struct makes sense

sumo-drosiek: See the latest code, please

docs/operators/file_input.md
sumo-drosiek: I updated documentation

djaglowski (approved):

djaglowski:
Thanks @sumo-drosiek, nice improvement!

PR #218, opened 2021-07-14
Add fieldalignment tool and check

pmalek-sumo:
Tracking issue: https://github.com/open-telemetry/opentelemetry-log-collection/issues/98


djaglowski (approved):
This is great. I confirmed it works locally.

What do you envision as far as managing the directories that it covers? Is there a technical reason not to include all directories? If not, should we do so?

pmalek-sumo:
I don't have a strong opinion on this. We can either take small steps or go for `./...` in this PR.

If this gets a 🟢 light then I can make the required changes proposed by `fieldalignment -fix ./...` and add those in this PR.

djaglowski:
Thanks @pmalek-sumo. I'll merge this, and we can enhance from here.

PR #246, opened 2021-08-25
Establish Autogenerated IDs for operators with unspecified IDs

### LINKED ISSUE #105: Autogenerated operator IDs

Mrod1598:
Resolves #105


pipeline/config.go
```diff
@@ -50,6 +53,32 @@ func (c Config) BuildOperators(bc operator.BuildContext, defaultOperator operato
...
 }
 
+func dedeplucateIDs(ops []operator.Config) {
+	typeMap := make(map[string]int)
+	for _, op := range ops {
+		if op.Type() != op.ID() {
+			continue
+		}
+
+		if typeMap[op.Type()] == 0 {
```
djaglowski, line 63: A nuance here, which I think is slightly less than idea, but probably ok:

I think this will "steal" a valid ID from an operator that hasn't been covered yet.
```
- copy
- copy
- copy1
```
would ideally become
```
- copy
- copy2
- copy1
```
but I believe it will actually be
```
- copy
- copy1
- copy2
```

I suggest we accept this behavior, but we should put a test case in place that ensures it does not change accidentally.

pipeline/config_test.go
```diff
@@ -0,0 +1,263 @@
...
+				ops = append(ops, operator.Config{Builder: newDummyOpConfig("json_parser", "json_parser")})
+				ops = append(ops, operator.Config{Builder: newDummyOpConfig("copy", "copy")})
+				ops = append(ops, operator.Config{Builder: newDummyOpConfig("json_parser1", "json_parser")})
+				ops = append(ops, operator.Config{Builder: newDummyOpConfig("copy1", "copy")})
+				ops = append(ops, operator.Config{Builder: newDummyOpConfig("json_parser2", "json_parser")})
+				return ops
+			}(),
+		},
+		{
+			"already_renamed",
```
djaglowski, line 130: I'd like to see some more variations on this one.

Cascading collisions:
```
- copy
- copy
- copy1
- copy2
- copy2
```


Out of order:
```
- copy2
- copy
- copy
- copy
```

As suggested above, "stealing ids"
```
- copy
- copy
- copy1
```

Can you think of others?

pipeline/config_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+
+type deduplicateTestCase struct {
+	name        string
+	ops         func() Config
+	expectedOps Config
+}
+
+func TestDeduplicateIDs(t *testing.T) {
```
djaglowski, line 42: Would it be reasonable to write another set of tests that validate how this affects plugins?

djaglowski (commented):
I like how simple this is looking to be. 

My biggest area of concern is with plugins. Have you determined whether they are affected by this, and if so how?

pipeline/config.go
Mrod1598: the second for loop just below this is for exactly that, it checks all the existing IDs to see if the new ID is already in use and increments the type if already exists. I'll add a test case for it.

pipeline/config_test.go
```diff
@@ -0,0 +1,263 @@
...
+	cases := []deduplicateTestCase{
+		{
+			"one_op_rename",
+			func() Config {
+				var ops Config
+				op1 := operator.Config{Builder: newDummyOpConfig("json_parser", "json_parser")}
+				ops = append(ops, op1)
+				op2 := operator.Config{Builder: newDummyOpConfig("json_parser", "json_parser")}
+				ops = append(ops, op2)
+				return ops
```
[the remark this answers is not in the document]
Mrod1598: reworked

pipeline/config_test.go
Mrod1598: won't that first part not even make it into deduplicate function as they have duplicate values?

pipeline/config.go
```diff
@@ -23,6 +25,7 @@ type Config []operator.Config
 
 // BuildOperators builds the operators from the list of configs into operators.
 func (c Config) BuildOperators(bc operator.BuildContext, defaultOperator operator.Operator) ([]operator.Operator, error) {
+	dedeplucateIDs(c)
```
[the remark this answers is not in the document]
Mrod1598: Updated

pipeline/config_test.go
Mrod1598: I have a set of tests within the `config_test` of plugin tests but I did just update them to remove the parts that were hardcoded in.

pipeline/config_test.go
djaglowski: Yep, you're right. The autogen functionality does not dedplicate all IDs, it just autogenerates IDs and makes sure they are not the same as other IDs.

djaglowski: I would still like to see a "already renamed but out of order" test:
```
"copy1"
"copy"
"copy"
```

testutil/operator_builder.go
```diff
@@ -63,6 +63,20 @@ func (_m *OperatorBuilder) ID() string {
 	return r0
 }
 
+// SetID provides a mock function with given fields: _a0
+func (_m *OperatorBuilder) SetID(_a0 string) error {
+	ret := _m.Called(_a0)
```
djaglowski, line 68: We're getting dinged on code coverage, but this is just test helper code. Let's rename this file to end with `_test.go` so it is excluded from coverage.

djaglowski (commented):
Looking good. Just a few more things.

pipeline/config_test.go
```diff
@@ -0,0 +1,278 @@
...
+				ops = append(ops, newDummyJSON("json_parser"))
+				ops = append(ops, newDummyJSON("json_parser1"))
+				ops = append(ops, newDummyJSON("json_parser2"))
+				ops = append(ops, newDummyJSON("json_parser3"))
+				ops = append(ops, newDummyJSON("json_parser4"))
+				return ops
+			}(),
+		},
+		{
+			"iterate_twice",
```
Mrod1598, line 146: @djaglowski I added one already does an out of order iteration here.

djaglowski: Fair enough. I would say that this test seems to cover more than the simplest form of this behavior, but it's fine.

testutil/operator_builder.go
Mrod1598: renamed this file

djaglowski (approved):
Nice work @Mrod1598!

PR #261, opened 2021-09-07
Update force flush default

### LINKED ISSUE #257: force_flush_period should default to non-zero value
See [related issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/4849)

Mrod1598:
Resolves #257

### DIFF
--- a/operator/builtin/input/file/config_test.go
+++ b/operator/builtin/input/file/config_test.go
@@ -706,7 +706,7 @@ func TestMapStructureDecodeConfigWithHook(t *testing.T) {
 			"line_start_pattern": expect.Splitter.Multiline.LineStartPattern,
 			"line_end_pattern":   expect.Splitter.Multiline.LineEndPattern,
 		},
-		"force_flush_period":   "500ms",
+		"force_flush_period":   0.5,
 		"include_file_name":    true,
 		"include_file_path":    false,
 		"start_at":             "end",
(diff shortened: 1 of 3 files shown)


operator/helper/multiline.go
```diff
@@ -33,7 +33,7 @@ type FlusherConfig struct {
 func NewFlusherConfig() FlusherConfig {
 	return FlusherConfig{
 		// Empty or `0s` means that we will never force flush
-		Period: Duration{Duration: 0},
+		Period: Duration{Duration: time.Hour},
```
djaglowski, line 36: Per the suggestion [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/4849#issuecomment-911926696), the default should be a few hundred milliseconds.

Since the default `poll_interval` is `200ms`, I suggest we use `500ms` here. This will ensure that at least two subsequent polls have occurred before flushing an incomplete log.

Also, we need to update the docs accordingly.

Mrod1598: whoops, must of misread and ok updated the doc as well.

operator/builtin/input/file/config_test.go
```diff
@@ -749,6 +750,9 @@ func TestMapStructureDecodeConfig(t *testing.T) {
 		"max_log_size":         1024 * 1024,
 		"max_concurrent_files": 1024,
 		"encoding":             "utf16",
+		"force_flush_period": map[string]interface{}{
+			"Duration": time.Millisecond * 500,
+		},
```
djaglowski, line 755: For the sake of consistency, let's make sure all duration parameters are instantiated the same way in this method. No reason to have `poll_interval` and `force_flush_period` instantiated differently.

djaglowski (approved):

PR #262, opened 2021-09-08
Update encoding noop

### LINKED ISSUE #138: Support byte stream data collection

Mrod1598:
Resolves #138

Added `MaxLogSize` to some of the build function args to allow use of it within the `SplitNone` function. this was to allow the function to respect max log size.


operator/helper/multiline_test.go
```diff
@@ -420,6 +421,90 @@ func TestNewlineSplitFunc(t *testing.T) {
...
+				"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuv",
+				"wxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqr",
+				"stuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmn",
+				"opqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkl",
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		splitFunc := SplitNone(100)
```
djaglowski, line 487: I would make this a constant and tie tests to it. Equal to, slightly greater, much greater, etc.

operator/builtin/input/file/file_test.go
```diff
@@ -205,6 +205,71 @@ func TestReadExistingLogs(t *testing.T) {
...
+	tcs := []struct {
+		testName string
+		input    []byte
+		test     func(*testing.T, chan *entry.Entry)
+	}{
+		{
+			"simple",
+			[]byte("testlog1"),
+			func(t *testing.T, c chan *entry.Entry) {
+				waitForMessage(t, c, "testlog1")
```
djaglowski, line 219: A really important part of this feature is that the resulting log body should be a `[]byte`.

Mrod1598: Updated to dump bytes on the emit func. Instead of decoding it, it just drops the bytes into the entry body.

operator/helper/multiline.go
```diff
@@ -204,6 +208,24 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
...
+	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
+		if len(data) >= maxLogSize {
+			return maxLogSize, data[:maxLogSize], nil
+		}
+
+		if !atEOF {
+			return 0, nil, nil
+		}
+
+		if atEOF && len(data) == 0 {
```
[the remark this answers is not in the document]
Mrod1598: fixed

operator/helper/multiline_test.go
Mrod1598: added more tests

djaglowski: This didn't really address what I was pointing out. The value `100` on this line should be a constant, because several of the tests depend on it. If we make it a constant, we can make the tests depend on the constant, which can then change without breaking tests.

operator/builtin/input/file/reader.go
```diff
@@ -172,7 +173,12 @@ func (r *Reader) emit(ctx context.Context, msgBuf []byte) error {
 		return fmt.Errorf("decode: %s", err)
 	}
 
-	e, err := r.fileInput.NewEntry(msg)
+	var e *entry.Entry
+	if r.fileInput.encoding.Encoding == encoding.Nop {
+		e, err = r.fileInput.NewEntry(msgBuf)
```
djaglowski, line 178: Since we aren't using `msg`, we should avoid computing it.

Mrod1598: updated

operator/helper/multiline_test.go
Mrod1598: added constant.

djaglowski (approved):

PR #303, opened 2021-11-09
Fix bug where splitting could be skipped

### LINKED ISSUE #292: Multiline splitter doesn't work all the time with default configuration
**Problem**:

If `force_flush_period` parameter is set to the default `500ms` or even a bit more, default multiline splitter occasionally doesn't work and pass log records with few original logs lines combined by `\n` to downstream operators. This causes failures in `regex_parser` operator.

**Steps to reproduce**:
1. Deploy some pods constantly generating logs in a k8s cluster. I used this demo app https://github.com/signalfx/microservices-demo
2. Deploy otel-collector with this sample configuration: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/examples/kubernetes/otel-collector.yaml + update the image version and change logging debug level to `info` to be able to easily notice the issue.

Usually within first couple of minutes, I can see errors from failing operators starting with `regex_parser`:

```
2021-10-25T00:44:35.396Z	error	Failed to process entry	{"kind": "receiver", "name": "filelog", "operator_id": "$.parser-containerd", "operator_type": "regex_parser", "error": "regex pattern does not match", "action": "send", "entry": {"timestamp":"2021-10-25T00:44:35.396252343Z","body":"2021-10-25T00:44:35.290698811Z stdout F {\"timestamp\": 1635122675.290467, \"severity\": \"INFO\", \"name\": \"recommendationservice-server\", \"message\": \"[Recv ListRecommendations] product_ids=['66VCHSJNUP', '9SIQT8TOJO', '0PUK6V6EV0', '1YMWWN1N4O', 'LS4PSXUNUM']\"}\n2021-10-25T00:44:35.361872731Z stdout F {\"timestamp\": 1635122675.3616176, \"severity\": \"INFO\", \"name\": \"recommendationservice-server\", \"message\": \"[Recv ListRecommendations] product_ids=['66VCHSJNUP', '2ZYFJ3GM2N', 'LS4PSXUNUM', 'L9ECAV7KIM', '6E92ZMYYFZ']\"}","attributes":{"file.path":"/var/log/pods/default_recommendationservice-54ff8b59dc-mz6z5_e6c4e740-15a1-4732-ba0a-a79b8846452f/server/0.log"},"severity":0}}
```

Not sure if it's related but seems like restarting the daemonset increases chances to catch the issue: `kubectl rollout restart ds otel-collector`.

Setting higher values for `force_flush_period` reduces chances to catch this. I got it with `500ms`, `1s` and `2s` and with default `poll_interval: 200ms`. There are no errors in the logs if `force_flush_period` is set to `0`.

I didn't play with logs volume, but I suspect increasing load can make it worse.

djaglowski:
Resolves #292 

When force_flush_period is set to a non-zero value,
a forced flushing operation would disregard the possibility
that the data being flushed may contain multiple entries.

For example, the file_input operator could flush the remainder
of a file that it was consuming. If multiple lines were written
to that file, they would be emitted as a single log entry.

The fix is basically that when a forced flush is required, we
will continue reading in lines until reaching the end of file.
Only then will we mark the flush complete.

### DIFF
--- a/operator/helper/multiline.go
+++ b/operator/helper/multiline.go
@@ -85,7 +85,7 @@ func (f *Flusher) CheckAndFlush() {
 }
 
 // ForceFlushed update struct fields after forced flush
-func (f *Flusher) ForceFlushed() {
+func (f *Flusher) Flush() {
 	f.force = false
 	f.lastForcedFlush = time.Now()
 }
@@ -156,7 +156,7 @@ func (c MultilineConfig) getSplitFunc(encodingVar encoding.Encoding, flushAtEOF 
 func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) bufio.SplitFunc {
 	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
 		if force != nil && force.ShouldFlush() {
-			force.ForceFlushed()
+			force.Flush()
 			token = trimWhitespaces(data)
 			advance = len(data)
 			return
@@ -231,7 +231,7 @@ func SplitNone(maxLogSize int) bufio.SplitFunc {
 func NewLineEndSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) bufio.SplitFunc {
 	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
 		if force != nil && force.ShouldFlush() {
-			force.ForceFlushed()
+			force.Flush()
 			token = trimWhitespaces(data)
 			advance = len(data)
 			return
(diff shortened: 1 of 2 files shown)


jsirianni (approved):
Initial testing looks good, I believe the issue is resolved. My process looks like this:
1. Deploy GKE cluster and replicate the issue using the [OTEL example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/examples/kubernetes/otel-collector.yaml) and the [Google Micro Service Demo](https://github.com/GoogleCloudPlatform/microservices-demo)
2. Observed the issue pretty quickly. Sometimes, logs would be appended to each other during a forced flush.
3. Verified that setting `force_flush_period` to `0` prevents the issue from happening
4. Reverted step 3, observed the issue again
5. Deployed this fix by building [Collector Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) against this branch
6. Verified that the issue is no longer happening

dmitryax (approved):
LGTM. Thanks for fixing it!

PR #304, opened 2021-11-12
Contribute Operators for Azure

jsirianni:
- Added Azure Event Hub Operator
- Added Azure Log Analytics Operator

Azure Event Hub can be leveraged to read arbitrary messages from Azures message queue service. Users can also configure Log Analytics to send logs to Event Hub, which can be read with the Log Analytics operator.


jsirianni:
Coverage is a bit low, I will work on increasing it for both operators. (see codecov ci failure).

go.mod
```diff
@@ -25,8 +25,28 @@ require (
 )
 
 require (
+	github.com/Azure/azure-event-hubs-go/v3 v3.3.16
+	golang.org/x/time v0.0.0-20210611083556-38a9dc6acbc6 // indirect
+	k8s.io/klog/v2 v2.9.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
+)
+
+require (
```
djaglowski, line 34: Will you run `go mod tidy`? Not clear why these are separate groups.

operator/builtin/input/azure/event_hub.go
```diff
@@ -0,0 +1,106 @@
...
+	azhub "github.com/Azure/azure-event-hubs-go/v3"
+	"go.uber.org/zap"
+)
+
+// Eventhub provides methods for reading events from Azure Event Hub.
+type EventHub struct {
+	AzureConfig
+	Persist *Persister
+	WG      sync.WaitGroup
+	Handler func(context.Context, *azhub.Event) error
```
djaglowski, line 31: Let's not export unless necessary.

operator/builtin/input/azure/event_hub_config.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	a.InputOperator = inputOperator
+
+	switch a.StartAt {
+	case "beginning":
+		a.startAtBeginning = true
+	case "end":
+		a.startAtBeginning = false
+	}
```
djaglowski, line 53: Should we have a default case that returns an error?

operator/builtin/input/azure/loganalytics/parse.go
```diff
@@ -0,0 +1,168 @@
...
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("failed to promote timestamp from %s field", key))
+				}
+				e.Timestamp = t
+				return nil
+			}
+		}
+	}
+	return nil
+}
```
djaglowski, line 156: This is super deep. Can we flatten?
```suggestion
func (l *LogAnalyticsInput) setTimestamp(e *entry.Entry, records map[string]interface{}) error {
	for key, value := range records {
		if key != "timegenerated" {
			continue
		}

		v, ok := value.(string)
		if !ok {
			continue
		}
		
		t, err := time.Parse("2006-01-02T15:04:05.0000000Z07", v)
		if err != nil {
			return errors.Wrap(err, fmt.Sprintf("failed to promote timestamp from %s field", key))
		}
				
		e.Timestamp = t
		return nil
	}
	return nil
}
```

operator/builtin/input/azure/loganalytics/parse.go
```diff
@@ -0,0 +1,168 @@
...
+				}
+
+				delete(records, key)
+				return records, nil
+			}
+			return nil, fmt.Errorf("expected '%s' field to be a string", typeField)
+		}
+	}
+	return nil, fmt.Errorf("expected to find field with name '%s'", typeField)
+}
```
djaglowski, line 139: Exit early to flatten

```suggestion
func (l *LogAnalyticsInput) setType(e *entry.Entry, records map[string]interface{}) (map[string]interface{}, error) {
	const typeField = "type"

	for key, value := range records {
		if key != typeField {
			continue
		}
		
		v, ok := value.(string)
		if !ok {
			return nil, fmt.Errorf("expected '%s' field to be a string", typeField)
		}
		
		v = strings.ToLower(v)

		// Set the log table label
		if err := l.setLabel(e, "azure_log_analytics_table", v); err != nil {
			return nil, err
		}

		delete(records, key)
		return records, nil
	}
	return nil, fmt.Errorf("expected to find field with name '%s'", typeField)
}
```

jsirianni: Much better, I commited this.

operator/builtin/input/azure/loganalytics/parse.go
jsirianni: Much better, I commited this.

operator/builtin/input/azure/event_hub_config.go
jsirianni: Done

operator/builtin/input/azure/event_hub.go
jsirianni: These are called by the operators in two seperate packages, `eventhub` and `loganalytics`.

go.mod
jsirianni: I think this is a Go 1.17 thing. `make tidy` does not change anything on my end.

PR #277, opened 2021-09-24
Contribute key value parser from observiq

jsirianni:
Key Value parser was originally implemented in the Stanza repo. The original PR has more context: https://github.com/observIQ/stanza/pull/426

### Example:

Input:
```
name=joe age=28 job=engineer t=1136214245 sev=info
```

Output:
```json
...
  "body": {
    "age": "28",
    "job": "engineer",
    "name": "joe"
  }
...
```

Issue #135, opened 2021-05-12
Parsers should not allow []byte type


djaglowski:
Parsers currently will cast `[]byte` into `string`, and emit a set of strings. Since protobuf does not allow non-utf8 bytes, we need to disallow this "backdoor" mechanism for converting `[]byte` to `string`, and force this conversion to take place as part of a decoding operation that can sanitize input.

This was discussed in Logs SIG today, and has additional context [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/3267).

PR #307, opened 2021-11-16
Kv parser

jsirianni:
Key Value parser was originally implemented in the Stanza repo. The original PR has more context: https://github.com/observIQ/stanza/pull/426

### Example:

Input:
```
name=joe age=28 job=engineer t=1136214245 sev=info
```

Output:
```json
...
  "body": {
    "age": "28",
    "job": "engineer",
    "name": "joe"
  }
...
```

Supersedes https://github.com/open-telemetry/opentelemetry-log-collection/pull/277. I lost my fork's branch, so I made a new one. 

I have implemented the delimiter feedback (documenting delimiter usage). I think the second delimiter feature could be added in a follow up PR.


operator/builtin/parser/keyvalue/keyvalue.go
```diff
@@ -0,0 +1,137 @@
...
+	return kv.ParserOperator.ProcessWith(ctx, entry, kv.parse)
+}
+
+// parse will parse a value as key values.
+func (kv *KVParser) parse(value interface{}) (interface{}, error) {
+	switch m := value.(type) {
+	case string:
+		return kv.parser(m, kv.delimiter)
+	case []byte:
+		return kv.parser(string(m), kv.delimiter)
```
djaglowski, line 84: This is no longer allowed as of #135

operator/builtin/parser/keyvalue/keyvalue.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (kv *KVParser) parser(input string, delimiter string) (map[string]interface{}, error) {
+	if len(input) == 0 {
+		return nil, fmt.Errorf("parse from field %s is empty", kv.ParseFrom.String())
+	}
+
+	parsed := make(map[string]interface{})
+
+	var err error
+	for _, raw := range splitStringByWhitespace(input) {
```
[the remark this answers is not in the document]
jsirianni: Done. The operator now takes a `func(input string) []string` for splitting the input string. By default, it will split on whitespace, if the user provides a character or substring, `stirngs.Split()` is used.

docs/operators/key_value_parser.md
```diff
@@ -0,0 +1,178 @@
...
+| `pair_delimiter` |                     | The delimiter used for seperating key value pairs, defaults to whitespace                                                                                                                                                                |
+| `output`         | Next in pipeline    | The connected operator(s) that will receive all outbound entries                                                                                                                                                                         |
+| `parse_from`     | $                   | A [field](/docs/types/field.md) that indicates the field to be parsed into key value pairs                                                                                                                                               |
+| `parse_to`       | $                   | A [field](/docs/types/field.md) that indicates the field to be parsed as into key value pairs                                                                                                                                            |
```
djaglowski, line 14: The library technically still interprets `$` as `$body`, but we've gotten away from documenting it this way. Better to be clear.

docs/operators/key_value_parser.md
```diff
@@ -0,0 +1,178 @@
+## `key_value_parser` operator
+
+The `key_value_parser` operator parses the string-type field selected by `parse_from` into key value pairs. All values are of type string.
+
+### Configuration Fields
+
+| Field            | Default             | Description                                                                                                                                                                                                                              |
```
djaglowski, line 7: Nit: To be consistent with other docs, can you add punctuation to the descriptions.

operator/builtin/parser/keyvalue/keyvalue.go
```diff
@@ -0,0 +1,167 @@
...
+func init() {
+	operator.Register("key_value_parser", func() operator.Builder { return NewKVParserConfig("") })
+}
+
+// NewKVParserConfig creates a new key value parser config with default values
+func NewKVParserConfig(operatorID string) *KVParserConfig {
+	return &KVParserConfig{
+		ParserConfig:  helper.NewParserConfig(operatorID, "key_value_parser"),
+		Delimiter:     "=",
+		PairDelimiter: "",
```
djaglowski, line 39:
```suggestion
```
Since this is the zero value for any string, not needed unless it adds clarity, but I don't think it does.

operator/builtin/parser/keyvalue/keyvalue.go
```diff
@@ -0,0 +1,167 @@
...
+		if len(m) != 2 {
+			e := fmt.Errorf("expected '%s' to split by '%s' into two items, got %d", raw, delimiter, len(m))
+			err = multierror.Append(err, e)
+			continue
+		}
+
+		key := cleanString(m[0])
+		value := cleanString(m[1])
+
+		// TODO: Check if key already exists and fail if so?
```
djaglowski, line 125: Failing entirely seems harsh. Probably a lot better to get a partial log than none. 

We could alternately log the error. Not sure the performance hit of checking every time is worth it for the rare case when it matters though.

Up to you, but we should decide and remove the TODO.

operator/builtin/parser/keyvalue/testdata/default.yaml
```diff
@@ -0,0 +1 @@
+type: key_value_parser
```
djaglowski, line 1:
```suggestion
type: key_value_parser

```

operator/builtin/parser/keyvalue/testdata/delimiter.yaml
```diff
@@ -0,0 +1,2 @@
+type: key_value_parser
+delimiter: ";"
```
djaglowski, line 2:
```suggestion
delimiter: ";"

```

djaglowski: We're not consistent on this in this library, but we need to move in that direction.

djaglowski:
Overall this looks good to me. Just a few minor cleanup things.

PR #315, opened 2021-11-26
feat(recombine): add `combine_with` option

### LINKED ISSUE #314: recombine operator: Add option to combine without newlines
Currently (v0.22.0) the [recombine](https://github.com/open-telemetry/opentelemetry-log-collection/tree/v0.22.0/operator/builtin/transformer/recombine) operator always stitches entries together using a newline character. This is described [here](https://github.com/open-telemetry/opentelemetry-log-collection/blob/v0.22.0/docs/operators/recombine.md?plain=1#L14):

> `combine_field`  | required         | The [field](/docs/types/field.md) from all the entries that will recombined with newlines.

and implemented [here](https://github.com/open-telemetry/opentelemetry-log-collection/blob/v0.22.0/operator/builtin/transformer/recombine/recombine.go#L247).

I propose to add an option to stitch the entries without the newline, or perhaps an option to change the stitching character/string to a different one (including none).

The reason for this proposal is to support stitching together the logs split by the Kubernetes CRI log format. For example, given this input file:

```
2016-10-06T00:17:09.669794202Z stdout P This is a very long line that has been spl
2016-10-06T00:17:09.669794202Z stdout F it according to the Kubernetes CRI format.
```

I want to be able to use the `recombine` operator to stitch the above two lines into one output line `This is a very long line that has been split according to the Kubernetes CRI format.`. Currently I'm not able to do this, instead what I'm getting is `This is a very long line that has been spl\nit according to the Kubernetes CRI format.`

This exact CRI scenario is in fact described in the operator's [documentation](https://github.com/open-telemetry/opentelemetry-log-collection/blob/v0.22.0/docs/operators/recombine.md?plain=1#L25-L73), but I believe the description is misleading. It suggests that the log lines with `P` tag and the log line with `F` tag should be combined as a single multiline log message, but I believe this is incorrect. These CRI log lines should be combined as a single one-line log message. In my experience, this is a common mistake to mix up these two concepts - stitching together one long line that has been split by the container runtime in multiple entries vs. combining multiple separate log lines into a single multiline log message. 

If there is another way to achieve my goal that I've missed, please kindly point me in the right direction. Otherwise, I'm happy to contribute this change as a pull request, including fixing the documentation.

astencel-sumo:
Fixes #314

### DIFF
--- a/docs/operators/recombine.md
+++ b/docs/operators/recombine.md
@@ -11,9 +11,10 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `on_error`       | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md). |
 | `is_first_entry` |                  | An [expression](/docs/types/expression.md) that returns true if the entry being processed is the first entry in a multiline series. |
 | `is_last_entry`  |                  | An [expression](/docs/types/expression.md) that returns true if the entry being processed is the last entry in a multiline series. |
-| `combine_field`  | required         | The [field](/docs/types/field.md) from all the entries that will recombined with newlines. |
+| `combine_field`  | required         | The [field](/docs/types/field.md) from all the entries that will recombined. |
+| `combine_with`   | `\n`             | The string that is put between the combined entries. This can be an empty string as well. |
 | `max_batch_size` | 1000             | The maximum number of consecutive entries that will be combined into a single entry. |
-| `overwrite_with` | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined with newlines. |
+| `overwrite_with` | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined. |
 
 Exactly one of `is_first_entry` and `is_last_entry` must be specified.
 
--- a/docs/operators/recombine.md
+++ b/docs/operators/recombine.md
@@ -22,40 +22,44 @@ NOTE: this operator is only designed to work with a single input. It does not ke
 
 ### Example Configurations
 
+#### Recombine Kubernetes logs in the CRI format
 
-#### Recombine logs in the CRI format
-
-Logs in the CRI format have a column that indicates whether the log is a partial log (P) or the last log in a series of partial logs (F). Using this column, we can recombine the CRI logs back into complete log messages.
+Kubernetes logs in the CRI format have a tag that indicates whether the log entry is part of a longer log line (P) or the final entry (F). Using this tag, we can recombine the CRI logs back into complete log lines.
 
 Configuration:
+
 ```yaml
 - type: file_input
-  include: 
+  include:
     - ./input.log
-- type: regex_parser 
-  regex: '^(?P<timestamp>[^\s]+) (?P<stream>\w+) (?P<partial>\w) (?P<message>.*)'
+- type: regex_parser
+  regex: '^(?P<timestamp>[^\s]+) (?P<stream>\w+) (?P<logtag>\w) (?P<message>.*)'
 - type: recombine
   combine_field: message
-  is_last_entry: "$body.partial == 'F'"
+  combine_with: ""
+  is_last_entry: "$body.logtag == 'F'"
+  overwrite_with: "newest"
 ```
 
-Input file: 
+Input file:
+
 ```
-2016-10-06T00:17:09.669794202Z stdout F The content of the log entry 1
-2016-10-06T00:17:10.113242941Z stdout P First line of log entry 2
-2016-10-06T00:17:10.113242941Z stdout P Second line of the log entry 2
-2016-10-06T00:17:10.113242941Z stdout F Last line of the log entry 2
+2016-10-06T00:17:09.669794202Z stdout F Single entry log 1
+2016-10-06T00:17:10.113242941Z stdout P This is a very very long line th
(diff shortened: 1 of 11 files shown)


pmalek-sumo (approved):
🚀

djaglowski:
@astencel-sumo This is a great improvement and I am fully in support of it.

That said, I think that we should have a test or two that demonstrate the configuration as it is read from an actual file, so that we can be sure there are not issues with the way a user would represent a newline character or any other character.

Would you mind adding a couple [golden config](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/builtin/transformer/move/config_test.go) tests to demonstrate the way a user would configure this?

astencel-sumo:
Thanks @djaglowski for pointing this out, will add the config tests.

Please not that for those to work, I will need this change: #317. Without it, I'm getting this failure:

```
$ go test
--- FAIL: TestConfig (0.00s)
    --- FAIL: TestConfig/default (0.00s)
        operatortest.go:85: 
                Error Trace:    operatortest.go:85
                                                        config_test.go:33
                Error:          Not equal: 
                                expected: &recombine.RecombineOperatorConfig{TransformerConfig:helper.TransformerConfig{WriterConfig:helper.WriterConfig{BasicConfig:helper.BasicConfig{OperatorID:"recombine", OperatorType:"metadata"}, OutputIDs:helper.OutputIDs(nil)}, OnError:"send", IfExpr:""}, IsFirstEntry:"", IsLastEntry:"", MaxBatchSize:1000, CombineField:entry.Field{FieldInterface:entry.FieldInterface(nil)}, CombineWith:"\n", OverwriteWith:"oldest"}
                                actual  : &recombine.RecombineOperatorConfig{TransformerConfig:helper.TransformerConfig{WriterConfig:helper.WriterConfig{BasicConfig:helper.BasicConfig{OperatorID:"recombine", OperatorType:"recombine"}, OutputIDs:helper.OutputIDs(nil)}, OnError:"send", IfExpr:""}, IsFirstEntry:"", IsLastEntry:"", MaxBatchSize:1000, CombineField:entry.Field{FieldInterface:entry.FieldInterface(nil)}, CombineWith:"\n", OverwriteWith:"oldest"}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -5,3 +5,3 @@
                                     OperatorID: (string) (len=9) "recombine",
                                -    OperatorType: (string) (len=8) "metadata"
                                +    OperatorType: (string) (len=9) "recombine"
                                    },
                Test:           TestConfig/default
FAIL
exit status 1
FAIL    github.com/open-telemetry/opentelemetry-log-collection/operator/builtin/transformer/recombine   1.438s
```

docs/operators/recombine.md
```diff
@@ -11,9 +11,10 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `on_error`       | `send`           | The behavior of the operator if it encounters an error. See [on_error](/docs/types/on_error.md). |
 | `is_first_entry` |                  | An [expression](/docs/types/expression.md) that returns true if the entry being processed is the first entry in a multiline series. |
 | `is_last_entry`  |                  | An [expression](/docs/types/expression.md) that returns true if the entry being processed is the last entry in a multiline series. |
-| `combine_field`  | required         | The [field](/docs/types/field.md) from all the entries that will recombined with newlines. |
+| `combine_field`  | required         | The [field](/docs/types/field.md) from all the entries that will recombined. |
+| `combine_with`   | `\n`             | The string that is put between the combined entries. This can be an empty string as well. |
```
djaglowski, line 15: Nit - since a user should expect different results based on whether or not this value is quoted, I think we need to be precise here. Specifically, the user would have to specify `"\n"` in order to replicate the default behavior, so I think we should show that exact value here.

djaglowski (commented):
Thanks for adding those config tests @astencel-sumo. Just one doc nit.

docs/operators/recombine.md
astencel-sumo: Yes I fully agree, we should be explicit about the types of the values.

djaglowski: Typically, it doesn't matter if the user writes `"foo"` or `foo`. When read, the value becomes the same either way.

In this case, it does matter whether the user writes `"\n"` or `\n`. I think it is enough to show the value here that is actually interpreted the same as the default value. 

```suggestion
| `combine_with`   | `"\n"`             | The string that is put between the combined entries. This can be an empty string as well. |
```

astencel-sumo: Fixed in https://github.com/open-telemetry/opentelemetry-log-collection/pull/315/commits/649bfc8a74dc505026d625f3135db78935b220a3.

Also updated examples in the docs, please take a look @djaglowski.

djaglowski (approved):
Thanks for the contribution @astencel-sumo. This is very nice improvement.

astencel-sumo:
Any plans for a new release @djaglowski? I'd like to use it in OTC's [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/61b560090fa1c0538409f578e0d011a89031e05a/receiver/filelogreceiver/go.mod#L9).

This change should probably also go into the [Changelog](https://github.com/open-telemetry/opentelemetry-log-collection/blob/v0.22.0/CHANGELOG.md), shouldn't it?

djaglowski:
Yes, it should go in the changelog. I'll make a PR to prep the changelog for the release, and will cut a release as soon as that's merged, most likely can be today.

PR #317, opened 2021-11-29
fix(recombine)!: change default operator type

astencel-sumo:
Here's a small inconsistency that I've found in the `recombine` operator. Can we fix it? Note that it is technically a breaking change.

When declaring a `recombine` operator in configuration without specifying the `id`, the default operator ID is `metadata`. I imagine this can cause issues when user configures both an actual `metadata` operator and `recombine` operator, both of them without their `id`s.

BREAKING CHANGE: The default operator ID for `recombine` operator is now `recombine`, not `metadata`.


djaglowski (approved):
Thanks for catching this.

PR #321, opened 2021-12-02
Update csv parser

JonathanWamsley:
- Added configuration support for Lazy Quotes
- Added configuration for a HeaderAttribute to support pulling header fields from an attribute on an entry. This is a pattern seen in W3C log format primarily for Microsoft IIS logging.
  - Due to headerAttribute creating generateParseFunc to dynamically create a parser function based on the HeaderAttribute
- Lower cased errors in csv to be in line with idiomatic Go

### DIFF
--- a/operator/builtin/parser/csv/csv.go
+++ b/operator/builtin/parser/csv/csv.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	csvparser "encoding/csv"
 	"fmt"
+	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


djaglowski:
@JonathanWamsley, these look like promising changes, but we'll need to process unrelated improvements as separate PRs. It looks like there should be at least 3 separate PRs here. Would you please split this up and open new PRs? I'll close this one for now.

PR #324, opened 2021-12-02
CSV Added 'Lazy Quotes' support

cpheps:
Added configuration for [Lazy Quotes](https://pkg.go.dev/encoding/csv#Reader) that passes through to the CSV reader.

Support a string like `"one",two,three` will parse as `"one"`, `two`, `three` with `lazy_quotes` enabled.


djaglowski:
Seems reasonable. Can you add the parameter to the documentation?

cpheps:
Yeah will do, overlooked that.

djaglowski (approved):

PR #335, opened 2021-12-20
Updated CSV Parser to include configurable header_attribute field.

cpheps:
Added configuration for a `header_attribute` in CSV parser to support pulling header fields from an attribute on an entry. This is a pattern seen in W3C log format primarily for Microsoft IIS logging.

When `header_attribute` is configured the csv parser will generate it parsing function at processing time. Refactored the parser logic to fit this need.

### DIFF
--- a/operator/builtin/parser/csv/csv.go
+++ b/operator/builtin/parser/csv/csv.go
@@ -103,6 +103,9 @@ type parseFunc func(interface{}) (interface{}, error)
 
 // Process will parse an entry for csv.
 func (r *CSVParser) Process(ctx context.Context, e *entry.Entry) error {
+	parse := r.parse
+
+	// If we have a headerAttribute set we need to dynamically generate our parser function
 	if r.headerAttribute != "" {
 		h, ok := e.Attributes[r.headerAttribute]
 		if !ok {
@@ -111,9 +114,10 @@ func (r *CSVParser) Process(ctx context.Context, e *entry.Entry) error {
 			return err
 		}
 		headers := strings.Split(h, string([]rune{r.fieldDelimiter}))
-		r.parse = generateParseFunc(headers, r.fieldDelimiter, r.lazyQuotes)
+		parse = generateParseFunc(headers, r.fieldDelimiter, r.lazyQuotes)
 	}
-	return r.ParserOperator.ProcessWith(ctx, e, r.parse)
+
+	return r.ParserOperator.ProcessWith(ctx, e, parse)
 }
 
 // generateParseFunc returns a parse function for a given header, allowing
--- a/operator/builtin/parser/csv/csv.go
+++ b/operator/builtin/parser/csv/csv.go
@@ -54,29 +54,27 @@ func (c CSVParserConfig) Build(context operator.BuildContext) ([]operator.Operat
 		return nil, err
 	}
 
-	if c.Header == "" && c.HeaderAttribute == "" {
-		return nil, errors.New("missing required field 'header' or 'header_attribute'")
-	}
-
-	if c.Header != "" && c.HeaderAttribute != "" {
-		return nil, errors.New("only one header parameter can be set: 'header' or 'header_attribute'")
-	}
-
 	if c.FieldDelimiter == "" {
 		c.FieldDelimiter = ","
 	}
+
+	fieldDelimiter := []rune(c.FieldDelimiter)[0]
 
 	if len([]rune(c.FieldDelimiter)) != 1 {
 		return nil, fmt.Errorf("invalid 'delimiter': '%s'", c.FieldDelimiter)
 	}
 
-	fieldDelimiter := []rune(c.FieldDelimiter)[0]
-
-	if c.HeaderAttribute == "" && !strings.Contains(c.Header, c.FieldDelimiter) {
+	headers := make([]string, 0)
+	switch {
+	case c.Header == "" && c.HeaderAttribute == "":
+		return nil, errors.New("missing required field 'header' or 'header_attribute'")
+	case c.Header != "" && c.HeaderAttribute != "":
+		return nil, errors.New("only one header parameter can be set: 'header' or 'header_attribute'")
+	case c.Header != "" && !strings.Contains(c.Header, c.FieldDelimiter):
 		return nil, errors.New("missing field delimiter in header")
+	case c.Header != "":
+		headers = strings.Split(c.Header, c.FieldDelimiter)
 	}
-
-	headers := strings.Split(c.Header, c.FieldDelimiter)
 
 	csvParser := &CSVParser{
 		ParserOperator:  parserOperator,
--- a/operator/builtin/parser/csv/csv.go
+++ b/operator/builtin/parser/csv/csv.go
@@ -18,6 +18,7 @@ import (
 	csvparser "encoding/csv"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/open-telemetry/opentelemetry-log-collection/entry"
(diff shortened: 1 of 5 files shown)


operator/builtin/parser/csv/csv.go
```diff
@@ -87,42 +94,66 @@ func (c CSVParserConfig) Build(context operator.BuildContext) ([]operator.Operat
...
+		h, ok := e.Attributes[r.headerAttribute]
+		if !ok {
+			err := fmt.Errorf("failed to read dynamic header attribute %s", r.headerAttribute)
+			r.Error(err)
+			return err
+		}
+		headers := strings.Split(h, string([]rune{r.fieldDelimiter}))
+		r.parse = generateParseFunc(headers, r.fieldDelimiter, r.lazyQuotes)
 	}
+	return r.ParserOperator.ProcessWith(ctx, e, r.parse)
```
djaglowski, line 118: Modifying the operator on the fly like this will limit the operator's ability to be used concurrently. Two different entries with different headers would potentially end up using one parse func. It doesn't look like there's any benefit to save the func anyways.

Also, maybe make `generateParseFunc` a receiver on the operator, since we're mostly just passing in a bunch of operator related config values anyways.

```suggestion
	if r.parse != nil {
		return r.ParserOperator.ProcessWith(ctx, e, r.parse)
	}
	
	h, ok := e.Attributes[r.headerAttribute]
	if !ok {
		err := fmt.Errorf("failed to read dynamic header attribute %s", r.headerAttribute)
		r.Error(err)
		return err
	}
	headers := strings.Split(h, string([]rune{r.fieldDelimiter}))
	parse := generateParseFunc(headers, r.fieldDelimiter, r.lazyQuotes)
	return r.ParserOperator.ProcessWith(ctx, e, parse)
```

operator/builtin/parser/csv/csv.go
```diff
@@ -87,42 +94,66 @@ func (c CSVParserConfig) Build(context operator.BuildContext) ([]operator.Operat
...
+			if err != nil {
+				return nil, err
+			}
+
+			for i, key := range headers {
+				parsedValues[key] = body[i]
+			}
+		}
+
+		return parsedValues, nil
```
djaglowski, line 157: It might be a little cleaner to seperate all the dynamic header related logic into a receiver on the operator. You don't have to pass a bunch of operator fields around, and can just generate, use, and forget the func all in one place.
```go
func (r *CSVParser) Process(ctx context.Context, e *entry.Entry) error {
	if r.parse != nil {
		return r.ParserOperator.ProcessWith(ctx, e, r.parse)
	}
	return r.dynamicParse(ctx, e)
}
```

```suggestion
func (r *CSVParser) dynamicParse(ctx context.Context, e *entry.Entry)  error {
	h, ok := e.Attributes[r.headerAttribute]
	if !ok {
		err := fmt.Errorf("failed to read dynamic header attribute %s", r.headerAttribute)
		r.Error(err)
		return err
	}
	headers := strings.Split(h, string([]rune{r.fieldDelimiter}))
	
	return r.ParserOperator.ProcessWith(ctx, e, 	
		func(value interface{}) (interface{}, error) {
			...
			reader := csvparser.NewReader(strings.NewReader(csvLine))
			reader.Comma = r.fieldDelimiter
			reader.FieldsPerRecord = len(headers)
			reader.LazyQuotes = r.lazyQuotes
			parsedValues := make(map[string]interface{})

			for {
				body, err := reader.Read()
				if err == io.EOF {
					break
				}

				if err != nil {
					return nil, err
				}

				for i, key := range headers {
					parsedValues[key] = body[i]
				}
			}

			return parsedValues, nil
		}
	)
```

operator/builtin/parser/csv/csv_test.go
```diff
@@ -197,14 +525,37 @@ func TestParserCSV(t *testing.T) {
...
+		// 		p.LazyQuotes = true
+		// 	},
+		// 	`stanza "log parser",1,6ft,5`,
+		// 	map[string]interface{}{
+		// 		"name":   `stanza "log parser"`,
+		// 		"age":    "1",
+		// 		"height": "6ft",
+		// 		"number": "5",
+		// 	},
+		// },
```
djaglowski, line 558: We can't have any commented out code. Does this need to be fixed or removed?

cpheps: Yeah that was a mistake forgot it was in there. Side effect of the rebase.

operator/builtin/parser/csv/csv.go
```diff
@@ -66,19 +72,20 @@ func (c CSVParserConfig) Build(context operator.BuildContext) ([]operator.Operat
 
 	fieldDelimiter := []rune(c.FieldDelimiter)[0]
 
-	if !strings.Contains(c.Header, c.FieldDelimiter) {
+	if c.HeaderAttribute == "" && !strings.Contains(c.Header, c.FieldDelimiter) {
```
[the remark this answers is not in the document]
cpheps: Going to switch the entire header check into a switch statement since it's more idiomatic than several if statements. and I'll make this change.

operator/builtin/parser/csv/csv.go
```diff
@@ -66,19 +72,20 @@ func (c CSVParserConfig) Build(context operator.BuildContext) ([]operator.Operat
...
 		return nil, errors.New("missing field delimiter in header")
 	}
 
-	numFields := len(strings.Split(c.Header, c.FieldDelimiter))
+	headers := strings.Split(c.Header, c.FieldDelimiter)
```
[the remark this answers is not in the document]
cpheps: It is in the current logic to generate and empty slice but this can be refactored.

operator/builtin/parser/csv/csv.go
cpheps: I think that's fair to change it to a receiver on the struct itself. 

Can you elaborate on why this would limit the parsers ability to be used concurrently? Since a parse function is being generated per entry and the it seems like the `ParseOperator` should be able to handle a new parse function every time, it's not obvious to me if we're creating a race condition further down in the stack.

operator/builtin/parser/csv/csv.go
cpheps: I tried this but I'm not sure I like it. You're right we do use some fields from the struct itself but also it means we regenerate the parse function every time. For configurations that are not using dynamic parsing (which is likely most of the time) we're just wasting processing time recalculating the same function over and over for every entry.

It is a bit ugly to have generate parse take in fields we have on the struct but also this way we're optimizing for the more common case instead of the edge case.

operator/builtin/parser/csv/csv.go
djaglowski: > Can you elaborate on why this would limit the parsers ability to be used concurrently?

Ok, roughly the steps to process one entry:
* Split headers
* Generate parse func
* Save parse func onto operator
* Process entry with operator's parse func

So if we have two entries, A & B, processed this way concurrently, they can interleave like:
* [A] Split headers
* [B] Split headers
* [A] Generate parse func
* [A] Save parse func onto operator
* [B] Generate parse func
* [B] Save parse func onto operator  // operator's parse func is now that of entry B
* [A] Process entry with operator's parse func // parsed with wrong parse func
* [B] Process entry with operator's parse func

operator/builtin/parser/csv/csv.go
djaglowski: > For configurations that are not using dynamic parsing (which is likely most of the time) we're just wasting processing time recalculating the same function over and over for every entry.

Not sure why we would do that. I don't think I suggested to do it that way. This func would only be called in the same circumstances where we currently call `generateParseFunc`.

djaglowski: To clarify this earlier comment:
```go
func (r *CSVParser) Process(ctx context.Context, e *entry.Entry) error {

	// r.parse set in Build function if using constant headers
	if r.parse != nil {
		return r.ParserOperator.ProcessWith(ctx, e, r.parse)
	}

	// only runs in dynamic headers case
	return r.dynamicParse(ctx, e)
}
```

cpheps: Sorry, I think I was operating under the assumption that the above issue didn't cause a race condition. This makes more sense now with that and I agree with your change I'll update it.

operator/builtin/parser/csv/csv.go
cpheps: Oh yeah saving the parse function we shouldn't save it in the dynamic case. We should just pass it through. The saving it is an optimization for the non dynamic case. I'll change this. Good call.

operator/builtin/parser/csv/csv.go
cpheps: The only issue with this is the initialization of the `parse` field on the struct is basically the `generateParseFunc` function we have now. By condensing all the parse logic into a receiver on the struct It makes it so you need to have an anonymous function inside `dynamicParse` in order to generate your parsing function.

I would propose keeping `generateParseFunc` and mixing it with the above code where if `r.parse` is nil return `r.ParserOperator.ProcessWith(ctx, e, generateParseFunc)`. Cleaned up but that's the gist.

cpheps: I pushed up a small change that avoids saving the dynamic function. Let me know if you still feel strongly that it should go into a receiver.

djaglowski (approved):

PR #341, opened 2022-01-11
Source identifier

### LINKED ISSUE #276: filelog: Multiline merging mixes up logs from different files
this works fine. 
```
operators:
    - default: clean-up-log-record
      routes:
      - expr: ($$resource["k8s.namespace.name"]) matches ".*" && ($$resource["k8s.container.name"])
          == "loggen"
        output: .*_loggen
      - expr: ($$resource["k8s.container.name"]) == "loggen2"
        output: loggen2
      type: router
    - combine_field: log
      id: .*_loggen
      is_first_entry: '($.log) matches "num: \\d+0\\s"'
      output: clean-up-log-record
      type: recombine
    - combine_field: log
      id: loggen2
      is_first_entry: '($$.log) matches "num: \\d+0\\s"'
      output: clean-up-log-record
      type: recombine
    - id: clean-up-log-record
      ops:
      - move:
          from: log
          to: $
      type: restructure
```
because `loggen` container and `loggen2` container is separately routed to its own `recombine` operator. 
but with configurations like this one
```
operators:
    - default: clean-up-log-record
      routes:
      - expr: ($$resource["k8s.namespace.name"]) matches ".*" && ($$resource["k8s.container.name"])
          == "loggen.*"
        output: .*_loggen.*
      type: router
    - combine_field: log
      id: .*_loggen.*
      is_first_entry: '($.log) matches "num: \\d+0\\s"'
      output: clean-up-log-record
      type: recombine
    - id: clean-up-log-record
      ops:
      - move:
          from: log
          to: $
      type: restructure
```
they get mixed up

rockb1017:
fixes #276 
Added sourceIdentifier parameter to prevent logs from multiple files being combined into one event.

Had to update the flush timeout logic to apply the timeout separately for each source.

### DIFF
--- a/docs/operators/recombine.md
+++ b/docs/operators/recombine.md
@@ -16,7 +16,7 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `max_batch_size`     | 1000             | The maximum number of consecutive entries that will be combined into a single entry. |
 | `overwrite_with`     | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined. |
 | `force_flush_period` | `5s`             | Flush timeout after which entries will be flushed aborting the wait for their sub parts to be merged with. |
-| `source_identifier`  |                  | The [field](/docs/types/field.md) to separate one source of logs from others when combining them. (i.e., `$attributes["file.path"]`) |
+| `source_identifier`  | `$attibutes["file.path"]` | The [field](/docs/types/field.md) to separate one source of logs from others when combining them. |
 
 Exactly one of `is_first_entry` and `is_last_entry` must be specified.
 
--- a/docs/operators/recombine.md
+++ b/docs/operators/recombine.md
@@ -16,6 +16,7 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `max_batch_size`     | 1000             | The maximum number of consecutive entries that will be combined into a single entry. |
 | `overwrite_with`     | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined. |
 | `force_flush_period` | `5s`             | Flush timeout after which entries will be flushed aborting the wait for their sub parts to be merged with. |
+| `source_identifier`  |                  | The [field](/docs/types/field.md) to separate one source of logs from others when combining them. (i.e., `$attributes["file.path"]`) |
 
 Exactly one of `is_first_entry` and `is_last_entry` must be specified.
 
--- a/docs/operators/recombine.md
+++ b/docs/operators/recombine.md
@@ -17,6 +17,7 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `overwrite_with`     | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined. |
 | `force_flush_period` | `5s`             | Flush timeout after which entries will be flushed aborting the wait for their sub parts to be merged with. |
 | `source_identifier`  | `$attibutes["file.path"]` | The [field](/docs/types/field.md) to separate one source of logs from others when combining them. |
+| `max_sources`        | 1000             | The maximum number of unique sources allowed concurrently to be tracked for combining separately. |
 
 Exactly one of `is_first_entry` and `is_last_entry` must be specified.
 
--- a/operator/builtin/transformer/recombine/recombine.go
+++ b/operator/builtin/transformer/recombine/recombine.go
@@ -156,11 +156,11 @@ func (r *RecombineOperator) flushLoop() {
 			for source, entries := range r.batchMap {
 				lastEntryTs := entries[len(entries)-1].Timestamp
 				timeSinceLastEntry := timeNow.Sub(lastEntryTs)
-				if timeSinceLastEntry > r.forceFlushTimeout {
-					err := r.flushSource(source)
-					if err != nil {
-						r.Errorf("there was error flushing combined logs %s", err)
-					}
+				if timeSinceLastEntry < r.forceFlushTimeout {
+					continue
+				}
+				if err := r.flushSource(source); err != nil {
+					r.Errorf("there was error flushing combined logs %s", err)
 				}
 			}
 
--- a/operator/builtin/transformer/recombine/recombine.go
+++ b/operator/builtin/transformer/recombine/recombine.go
@@ -262,7 +262,7 @@ func (r *RecombineOperator) matchIndicatesLast() bool {
 func (r *RecombineOperator) addToBatch(ctx context.Context, e *entry.Entry, source string) {
 	if _, ok := r.batchMap[source]; !ok {
 		r.batchMap[source] = []*entry.Entry{e}
-		if len(r.batchMap) > r.maxSources {
+		if len(r.batchMap) >= r.maxSources {
 			r.Error("Batched source exceeds max source size. Flushing all batched logs. Consider increasing max_sources parameter")
 			r.flushUncombined(ctx)
 		}
--- a/operator/builtin/transformer/recombine/recombine.go
+++ b/operator/builtin/transformer/recombine/recombine.go
@@ -52,6 +52,7 @@ type RecombineOperatorConfig struct {
 	MaxBatchSize             int           `json:"max_batch_size"     yaml:"max_batch_size"`
 	CombineField             entry.Field   `json:"combine_field"      yaml:"combine_field"`
 	CombineWith              string        `json:"combine_with"       yaml:"combine_with"`
+	SourceIdentifier         entry.Field   `json:"source_identifier"  yaml:"source_identifier"`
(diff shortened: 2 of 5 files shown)


docs/operators/recombine.md
```diff
@@ -16,6 +16,7 @@ The `recombine` operator combines consecutive logs into single logs based on sim
 | `max_batch_size`     | 1000             | The maximum number of consecutive entries that will be combined into a single entry. |
 | `overwrite_with`     | `oldest`         | Whether to use the fields from the `oldest` or the `newest` entry for all the fields that are not combined. |
 | `force_flush_period` | `5s`             | Flush timeout after which entries will be flushed aborting the wait for their sub parts to be merged with. |
+| `source_identifier`  |                  | The [field](/docs/types/field.md) to separate one source of logs from others when combining them. (i.e., `$attributes["file.path"]`) |
```
djaglowski, line 19: I would support having `$attributes["file.path"]` as the default value here. It's a value that is produced by this library, so it's not unreasonable to expect it. 

Having a default means that we are _always_ going to manage recombined lines by source. I can't think of any scenarios where this wouldn't be appropriate behavior (i.e. how would one reasonably expect recombination from multiple sources to work?).

operator/builtin/transformer/recombine/recombine.go
```diff
@@ -147,9 +152,19 @@ func (r *RecombineOperator) flushLoop() {
...
+			timeNow := time.Now()
+			for source, entries := range r.batchMap {
+				lastEntryTs := entries[len(entries)-1].Timestamp
+				timeSinceLastEntry := timeNow.Sub(lastEntryTs)
+				if timeSinceLastEntry > r.forceFlushTimeout {
+					err := r.flushSource(source)
+					if err != nil {
+						r.Errorf("there was error flushing combined logs %s", err)
+					}
+				}
```
djaglowski, line 164: Nit: This is quite deeply nested. Perhaps slightly more idiomatic:

```suggestion
			timeNow := time.Now()
			for source, entries := range r.batchMap {
				lastEntryTs := entries[len(entries)-1].Timestamp
				timeSinceLastEntry := timeNow.Sub(lastEntryTs)
				if timeSinceLastEntry < r.forceFlushTimeout {
					continue
				}
				if err := r.flushSource(source); err != nil {
					r.Errorf("there was error flushing combined logs %s", err)
				}
```

operator/builtin/transformer/recombine/recombine.go
```diff
@@ -190,11 +205,20 @@ func (r *RecombineOperator) Process(ctx context.Context, e *entry.Entry) error {
...
+	var s string
+	if r.sourceIdentifier.FieldInterface != nil {
+		err := e.Read(r.sourceIdentifier, &s)
+		if err != nil {
+			r.Warn("entry does not contain the source_identifier, so it may be pooled with other sources")
+			s = "DefaultSourceIdentifier"
+		}
+	} else {
+		s = "DefaultSourceIdentifier"
+	}
```
djaglowski, line 217: We can simplify some of this by having a default setting (suggested above).

operator/builtin/transformer/recombine/recombine.go
```diff
@@ -230,45 +254,66 @@ func (r *RecombineOperator) matchIndicatesLast() bool {
...
+	if r.batchSize >= r.maxBatchSize {
```
djaglowski, line 268: The way you've chosen to use this is interesting. We are batching entries by source, but counting them all together as an overall maximum. My first instinct was to suggest that we just allow each batch to hit a max size, but the overall max has a couple nice benefits for memory usage. For one, it roughly limits the overall size of memory that may be need to be allocated by the operator. The other benefit is that it provides a simple way to garbage collect old sources over time. 

All that said, I wonder if this could lead to difficult-to-explain behavior in some cases. For example, as a user, I think I would expect that `max_batch_size` will only place a per-source limit on the number of entries that would be combined into a single message. Let's say I know my system splits all logs into two parts, so I just set `max_batch_size: 2`. Now if I have more than one source, I will have lots of premature flushes.

An alternate design, which I think would solve all the same problems while being a little easier to understand:
- Let's apply `max_batch_size` to each source individually.
  - This allows us to remove `r.batchSize` and just depend on `len(batchMap[s])`
- Add another config setting called `max_sources` with a reasonable default, maybe 1000
  - At the end of `addToBatch`, check `if len(batchMap) > r.maxSources`
    - If so, we need to determine flushing logic. I think it's probably ok to just flush all at this point
- Whenever we flush entries for a source, `delete(batchMap, s)` to ensure we are tracking number of sources correctly

djaglowski (commented):
@rockb1017 Thanks for working on this. I like how it's looking but have some suggestions on one aspect of the design. Curious if you agree.

docs/operators/recombine.md
rockb1017: agree, i will update it.

operator/builtin/transformer/recombine/recombine.go
```diff
@@ -229,46 +255,61 @@ func (r *RecombineOperator) matchIndicatesLast() bool {
...
+	if _, ok := r.batchMap[source]; !ok {
+		r.batchMap[source] = []*entry.Entry{e}
+	} else {
+		r.batchMap[source] = append(r.batchMap[source], e)
+	}
+
+	if len(r.batchMap) > r.maxSources {
+		r.Error("Batched source exceeds max source size. Flushing all batched logs. Consider increasing max_sources parameter")
+		r.flushUncombined(ctx)
+	}
```
djaglowski, line 276: There are a couple opportunities to improve performance here, just by reorganizing the code.
1. We only need to check `if len(r.batchMap[source]) > r.maxBatchSize` after adding to an existing batch. Also, by checking this immediately _after_ adding to a batch, we will flush as soon as we hit the limit, rather than wait until next cycle.
2. We only need to check `if len(r.batchMap) > r.maxSources` after initializing a new batch.

Something like:
```go
if _, ok := r.batchMap[source]; !ok {
	r.batchMap[source] = []*entry.Entry{e}
	if len(r.batchMap) > r.maxSources {
		r.Error("Batched source exceeds max source size. Flushing all batched logs. Consider increasing max_sources parameter")
		r.flushUncombined(ctx)
	}
	return
}

r.batchMap[source] = append(r.batchMap[source], e)
if len(r.batchMap[source]) > r.maxBatchSize {	
	if err := r.flushSource(ctx, source); err != nil {
		r.Errorf("there was error flushing combined logs %s", err)
	}
}
```

operator/builtin/transformer/recombine/recombine_test.go
```diff
@@ -141,6 +149,27 @@ func TestRecombineOperator(t *testing.T) {
...
+				entryWithBodyAttr(t1, "file1", map[string]string{"file.path": "file1"}),
+				entryWithBodyAttr(t1, "file2", map[string]string{"file.path": "file2"}),
+				entryWithBodyAttr(t2, "end", map[string]string{"file.path": "file1"}),
+				entryWithBodyAttr(t2, "end", map[string]string{"file.path": "file2"}),
+			},
+			[]*entry.Entry{
+				entryWithBodyAttr(t1, "file1\nend", map[string]string{"file.path": "file1"}),
+				entryWithBodyAttr(t1, "file2\nend", map[string]string{"file.path": "file2"}),
+			},
+		},
```
djaglowski, line 172: Would you be willing to add a couple more test cases? 
- That `DefaultSourceIdentifier` works as expected
- That `max_sources` works as expected
- That `max_batch_size` works as expected (Not part of your changes, but would be appreciated since it's related)

rockb1017: sounds good. added.

operator/builtin/transformer/recombine/recombine.go
rockb1017: awesome find!

rockb1017:
hmm i got this `fatal error: concurrent map writes` during bench test. ` go test -benchmem -bench .`
any insight you can share?

operator/builtin/transformer/recombine/recombine.go
```diff
@@ -229,46 +259,59 @@ func (r *RecombineOperator) matchIndicatesLast() bool {
...
 // flushUncombined flushes all the logs in the batch individually to the
 // next output in the pipeline. This is only used when there is an error
 // or at shutdown to avoid dropping the logs.
 func (r *RecombineOperator) flushUncombined(ctx context.Context) {
-	for _, entry := range r.batch {
-		r.Write(ctx, entry)
+	for source := range r.batchMap {
+		for _, entry := range r.batchMap[source] {
+			r.Write(ctx, entry)
+		}
```
djaglowski, line 287: It would be nice if we combined entries for each source, but given that this is a fairly atypical bailout condition, I think it's ok to proceed with individual entries. I'll make a ticket to capture this as a future improvement.

djaglowski (approved):

PR #352, opened 2022-01-20
Bump gosec

djaglowski:
`make install-tools` was failing unless both packages were pinned, and in this order.


jsirianni (approved):
Working great on my machine :+1:

PR #355, opened 2022-01-21
Give recombine timeout test a larger margin for error

### LINKED ISSUE #354: Recombine operator tests fail on Windows
I noticed this issue while reviewing a PR: https://github.com/open-telemetry/opentelemetry-log-collection/runs/4899215817?check_suite_focus=true

flaky_windows_test

djaglowski:
This test is meant to ensure that partial logs are flushed
on an interval if they are not 'completed'. The test
appears to have too unforgiving a margin when ensuring that
the timeout is not applied too soon.

Resolves #354


djaglowski:
I ran CI twice w/o a failure, but I think this was just an unlikely race condition so probably we won't see it very easily.

jsirianni (approved):

Issue #164, opened 2021-05-26
"Field" syntax is consumed by OTel collector config loading process


djaglowski:
The OTel collector [automatically expands](https://github.com/open-telemetry/opentelemetry-collector/blob/df76aa36cd124ba4ae1b3d1ac1941155f86ea568/config/configloader/config.go#L515) tokens that appear to be environment variables. (i.e. `$hostname`)

This library's [field](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/docs/types/field.md#fields) syntax relies heavily on usage of `$resources`, `$attributes`, and `$body`. 

The documented approach to specifying fields does not work. Instead, the `$resources`, `$attributes`, and `$body` tokens are expanded (typically replaced with an empty string).

It is possible to escape the expansion (`$$attributes`), but as a component of the open telemetry project, we would ideally not require users to worry about this.

Fortunately, this usage of `$` in this library was a largely arbitrary decision. I suggest that we should choose an alternate character or mechanism for field syntax.

pmalek-sumo:
I wanted to open an new issue but I thought this is connected somehow:

So my config:

```yaml
  filelog:
    include: [ "**.log" ]
    operators:
      - type: metadata
        id: metadata1
        resource:
          ttt: 'EXPR( $ == "x" ? "asd" : "qwe" )'
      - type: metadata
        id: metadata2
        resource:
          resource2: $resource
      - type: metadata
        id: metadata3
        attributes:
          attribute2: $attributes
```

producing (empty resource and attributes) entries:

```
2021-07-13T18:50:21.161+0200    INFO    loggingexporter/logging_exporter.go:135 LogsExporter    {"#logs": 1}
2021-07-13T18:50:21.161+0200    DEBUG   loggingexporter/logging_exporter.go:141 ResourceLog #0
Resource labels:
     -> ttt: STRING(qwe)
     -> resource2: STRING()
InstrumentationLibraryLogs #0
InstrumentationLibrary
LogRecord #0
Timestamp: 2021-07-13 16:50:21.062134 +0000 UTC
Severity: Undefined
ShortName:
Body: 2021-07-13T18:50:20.995+0200      debug   sumologicextension@v0.27.0/extension.go:281     Heartbeat sent  {"kind": "extension", "name": "sumologic/1"}
Attributes:
     -> file.name: STRING(reg.log)
     -> attribute2: STRING()
```

[...]

PR #364, opened 2022-02-02
Remove '$' from field syntax

djaglowski:
This is a possible resolution to #164. This would be a major breaking change, so if accepted, the timing of a merge should be considered carefully.

---

This library has a construct called 'field', which has the
purpose of providing a concise way to refer to values in a
log record. The special symbol `$` has been used in key
words that refer to top level fields in the log record.
i.e. `$body`, `$attributes`, `$resource`

Additionally, some shorthand notion was supported. Notably,
`$.foo` and `.foo` were equivalent to `$body.foo`.

This change proposes to remove the usage of `$` in field syntax.
The main idea here is that all usages of field syntax MUST
begin with a key word i.e. `body`, `attributes`, `resource`.
With this requirement built in, there is no need to differentiate
between top level fields and arbitrary keys. The distinction is
implicit because the first word is always a top level field and
the remainder are arbitrary nested keys.


djaglowski:
@jsirianni, @pmm-sumo, @rockb1017, please take a look. Any reservations on this approach?

tigrannajaryan:
@dmitryax please have a look. Will our helm chart be affected by this change?

tigrannajaryan:
The change looks good to me. The syntax becomes clearer.

I do not see any downsides to this change, are there any?

pmm-sumo (approved):

pmalek-sumo (commented):
I like the idea of making it more explicit and easier to reason about.

🚀 some nits in `.md` tables

dmitryax:
> @dmitryax please have a look. Will our helm chart be affected by this change?

It will be affected, but we can adopt for this change.

I like the proposal. The syntax seems cleaner now

dmitryax (approved):

rockb1017 (approved):

dmitryax (commented):
LGTM, but should we throw deprecation log warnings?

djaglowski:
> LGTM, but shouldn't we throw deprecation warnings in logs?

Unfortunately, there's not a great way to complain about the deprecated usage because it is processed during unmarshaling, where we don't have a handle to a logger. We could add a flag to the `Field` struct, set it during unmarshaling, and then log every time the field is used in the pipeline, but this seems overkill to me. In my opinion, logging about deprecations is a nice-to-have feature and probably not worth any substantial code changes to do it. If you feel strongly though, I can make the above change, or perhaps you see a better way that I missed. 

Another option is to fully remove support, but to detect the old syntax and return a clear error. This forces migration immediately, but at least would not be ambiguous at all.

dmitryax:
Maybe we can drop the old interface support here but do all the necessary translation from the deprecated format in the collector-contrib? There we can throw all the warning during translation.

djaglowski:
Unfortunately, I don't think unmarshaling in collector-contrib gives us a better opportunity to do this.

djaglowski:
@dmitryax, can you clarify whether you see the lack of deprecation warning as a blocker? 

I do not see a good way to log warnings here or at the collector level, so I would prefer to either error or quietly support the old syntax.

dmitryax:
What if we return an error of a `DeprecationError` type from unmarshal functions? Then contrib can look for that error and throw a warning message to the output with everything else working as usual.

djaglowski:
> What if we return an error of a `DeprecationError` type from unmarshal functions? Then contrib can look for that error and throw a warning message to the output with everything else working as usual.

I liked this idea and gave it a shot, but unfortunately, the `yaml` library does not appear to allow for partial failures like this. Any error returned ultimately triggers a call to [fail](https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/yaml.go#L259).

dmitryax:
I see, if it's not possible to show a warning, I don't think silently support both versions is useful. Changing the interface right away seems to be a better option then.

dmitryax (approved):

PR #370, opened 2022-02-10
Add ObservedTimestamp to entry

### LINKED ISSUE #367: Add ObservedTimestamp Field
The notion of an Observed Timestamp is a [recent addition to the Logs Data Model](https://github.com/open-telemetry/opentelemetry-specification/pull/2184).

This field should be added to the library and populated automatically as appropriate.
- [ ] Add ObservedTimestamp field to entry.Entry
- [ ] entry.New should initialize ObservedTimestamp, but not Timestamp
- [ ] All usages of Timestamp must be evaluated. (i.e. what happens if not set?)
- [x] Ensure all preexisting assignments to Timestamp are still valid (excepting entry.New initialization)

djaglowski:
Resolves #367 

This is a breaking change because `entry.Timestamp` is no longer initialized by default.

### DIFF
--- a/entry/entry_test.go
+++ b/entry/entry_test.go
@@ -133,9 +133,12 @@ func TestRead(t *testing.T) {
 }
 
 func TestCopy(t *testing.T) {
+	now := time.Now()
+
 	entry := New()
 	entry.Severity = Severity(0)
 	entry.SeverityText = "ok"
+	entry.ObservedTimestamp = now
 	entry.Timestamp = time.Time{}
 	entry.Body = "test"
 	entry.Attributes = map[string]interface{}{"label": "value"}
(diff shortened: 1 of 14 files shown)


tigrannajaryan (commented):
Just to confirm: any existing parsing operation will continue to write to `Timestamp`. If there is no parsing than the `Timestamp` will remain empty. `ObservedTimestamp` will be always initialized to `now` and is not affected by parsers. Is this correct?

djaglowski:
> Just to confirm: any existing parsing operation will continue to write to `Timestamp`. If there is no parsing than the `Timestamp` will remain empty. `ObservedTimestamp` will be always initialized to `now` and is not affected by parsers. Is this correct?

That's exactly correct.

jsirianni (approved):

entry/entry_test.go
```diff
@@ -133,9 +133,12 @@ func TestRead(t *testing.T) {
 }
 
 func TestCopy(t *testing.T) {
+	now := time.Now()
```
jsirianni, line 136: To be consistent:
```suggestion
	now := timeNow()
```

djaglowski: I disagree on this one. `timeNow` only exists for when a timestamp cannot be otherwise controlled. 

Specifically, `entry.New` must set `ObservedTimestamp`. In order to test that this is done, we need to know ahead of time what that timestamp should be. So, `entry.New` calls `timeNow`, which is `time.Now` in production, but can be overridden for specific test cases where the timestamp cannot otherwise be controlled.

Issue #375, opened 2022-02-10
Remove plugin package and related complexity


djaglowski:
This library supports a concept of plugins that is unused in the OTel project. The implementation of these plugins drastically complicates the code that builds pipelines and operators. I believe it is time to remove plugins from this repo and reduce the complexity of related code as allowed by this change. Because OTel eventually will implement a similar concept at the collector level, it's worth noting that this code will still be available for reference/usage in the git history.

PR #376, opened 2022-02-10
Syslog input as wrapper

djaglowski:
The syslog_input operator was previously implemented
as a builder that built two operators. While this is
a perfectly fine solution, it happens to be the only
place in this codebase where multiple operators are
returned from a builder, excepting plugins.

This new implementation builds the same two operators,
but then wraps them together within a single operator.
This will allow for the eventual simplification of
other code, once plugins are also removed from this
codebase.

Related to #375, #380

### DIFF
--- a/operator/builtin/input/syslog/syslog.go
+++ b/operator/builtin/input/syslog/syslog.go
@@ -41,19 +41,15 @@ type SyslogInputConfig struct {
 }
 
 func (c SyslogInputConfig) Build(context operator.BuildContext) ([]operator.Operator, error) {
+
 	inputBase, err := c.InputConfig.Build(context)
 	if err != nil {
 		return nil, err
 	}
 
-	parentID := c.InputConfig.ID()
-	if parentID == "" {
-		parentID = c.InputConfig.Type()
-	}
-
-	syslogParserCfg := syslog.NewSyslogParserConfig(parentID + "_internal_tcp")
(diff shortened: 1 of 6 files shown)


djaglowski:
The contrib failure is expected due to this change. I've proven out a change to `syslogreceiver.DecodeInputConfig` that will fix it:

```go
// DecodeInputConfig unmarshals the input operator
func (f ReceiverType) DecodeInputConfig(cfg config.Receiver) (*operator.Config, error) {
	logConfig := cfg.(*SysLogConfig)
	yamlBytes, _ := yaml.Marshal(logConfig.Input)
	inputCfg := syslog.NewSyslogInputConfig("syslog_input")
	inputCfg.SyslogBaseConfig = syslogparser.NewSyslogParserConfig("syslog_parser").SyslogBaseConfig

	if err := yaml.Unmarshal(yamlBytes, &inputCfg); err != nil {
		return nil, err
	}
	return &operator.Config{Builder: inputCfg}, nil
}
```

djaglowski:
This change ended up simplifying the `syslog_input` operator in some notable ways. Particularly, I like that it removed the need for a custom unmarshal function.

jsirianni:
@djaglowski do we expect contrib-tests to fail?

djaglowski:
> @djaglowski do we expect contrib-tests to fail?

Yeah, see comment above.

jsirianni (approved):
I built this against my local contrib and tested TCP and UDP against rfc5424, looks good.

PR #377, opened 2022-02-13
Remove plugins

### LINKED ISSUE #375: Remove plugin package and related complexity

djaglowski:
Plugins are an unused feature of this library. They will likely be
reimplemented at the collector level.

Resolves #375


djaglowski:
The `plugin` package had a high level of test coverage, so removing it will bring down the average on the repo. I think this is not a reason to block this change, nor should increased coverage in unrelated packages be required here.

tigrannajaryan (approved):

PR #386, opened 2022-02-14
Make operator.Builder.Build return one operator

### LINKED ISSUE #380: operator.Builder.Build should return only a single operator
The `operator.Builder` interface currently expects the `Build` function to return a slice of operators. This was necessary to support the notion of plugins, and was taken advantage of in the `syslog_input` operator as well. However, plugins are being removed in #377 and `syslog_input` has been simplified in #376. We should allow callers of this function to expect only a single operator. This will allow for the removal of quite a bit of unnecessary "get the first element" code throughout the codebase.

djaglowski:
The `operator.Builder.Build` function previously required support
for building multiple operators. This was necessary for plugins
and convenient in one other case, but generally just complicated the 
codebase, as many callers had to drill work with a slice even when 
knowing that there would only be one operator in it.

Resolves #380

### DIFF
--- a/operator/builtin/input/file/benchmark_test.go
+++ b/operator/builtin/input/file/benchmark_test.go
@@ -156,9 +156,8 @@ func BenchmarkFileInput(b *testing.B) {
 			}
 			cfg.StartAt = "beginning"
 
-			ops, err := cfg.Build(testutil.NewBuildContext(b))
+			op, err := cfg.Build(testutil.NewBuildContext(b))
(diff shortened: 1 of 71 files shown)


djaglowski:
The contrib test failure is expected. [Here](https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/main...djaglowski:log-collection-update) is a diff showing the changes that will be required in order to update contrib.

jsirianni (approved):

PR #387, opened 2022-02-14
Remove notion of operator namespaces

### LINKED ISSUE #383: Remove operator ID prefixing
This issue follows from #375.

This library requires all operators to have unique identifiers. When building plugins, which contained multiple operators, it became necessary to manage operator id within a in a hierarchy. This required an entire system of build context and ID prefixing. Without plugins, there is really no need for all this complexity, so it can be safely removed.

djaglowski:
Namespaces were previously used to differentiate operators
within plugins or instances of plugins. The idea was to apply a
prefix to each operator id within a plugin. Nested plugins would have
multiple prefixes. In every case, the id would end up unique, barring
some extremely verbose manual naming (which would be caught at startup).
    
Many of these line changes are due to removal of the "root" namespace,
which was just a '$.' prefixed to each operator id.

Resolves #383

### DIFF
--- a/agent/builder_test.go
+++ b/agent/builder_test.go
@@ -70,22 +70,22 @@ func TestBuildAgentDefaultOperator(t *testing.T) {
 
 	for _, op := range ops {
 		switch op.ID() {
-		case "$.noop":
+		case "noop":
 			require.Equal(t, 1, len(op.GetOutputIDs()))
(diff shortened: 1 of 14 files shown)


djaglowski:
Contrib test failures are expected, but will be resolved after the next release using changes shown [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/main...djaglowski:log-collection-update).

PR #397, opened 2022-02-17
Provide a dedicated mechanism for parsing logger name

### LINKED ISSUE #390: Provide dedicated capability to isolate logger name
This change follows from [(specification) #2276](https://github.com/open-telemetry/opentelemetry-specification/pull/2276)

This library should provide an simple mechanism for isolating a LoggerName value, which will be used later in the data pipeline (specifically [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/293217ba697f56585bc4e9e700bae82d09be15c4/internal/stanza/converter.go#L245)), to populate the InstrumentationScope.

The following is suggested:
1. Add `LoggerName` to `entry.Entry`. (This does not imply that `LoggerName` is a top level field in the data model. It is simply a means to convey this information until the log entry is batched by Instrumentation Scope.)
2. Add a `logger_name` block, which follows the same pattern as `timestamp` and `severity`. Specifically, this means that every parser may optionally assign a string value to LoggerName.

djaglowski:
Resolves #390

### DIFF
--- a/docs/types/scope_name.md
+++ b/docs/types/scope_name.md
@@ -1,0 +1,64 @@
+## Scope Name Parsing
+
+A Scope Name may be parsed from a log entry in order to indicate the code from which a log was emitted.
+
+### `scope_name` parsing parameters
+
+Parser operators can parse a scope name and attach the resulting value to a log entry.
+
+| Field          | Default   | Description |
+| ---            | ---       | ---         |
+| `parse_from`   | required  | The [field](/docs/types/field.md) from which the value will be parsed. |
+| `preserve_to`  |           | Preserves the unparsed value at the specified [field](/docs/types/field.md). |
+
+
+### How to use `scope_name` parsing
+
+All parser operators, such as [`regex_parser`](/docs/operators/regex_parser.md) support these fields inside of a `scope_name` block.
+
+If a `scope_name` block is specified, the parser operator will perform the parsing _after_ performing its other parsing actions, but _before_ passing the entry to the specified output operator.
+
+
+### Example Configurations
+
+#### Parse a scope_name from a string
+
+Configuration:
+```yaml
+- type: regex_parser
+  regexp: '^(?P<scope_name_field>\S*)\s-\s(?P<message>.*)'
+  scope_name:
+    parse_from: body.scope_name_field
+```
+
+<table>
+<tr><td> Input entry </td> <td> Output entry </td></tr>
+<tr>
+<td>
+
+```json
+{
+  "resource": { },
+  "attributes": { },
+  "body": "com.example.Foo - some message",
+  "scope_name": "",
+}
+```
+
+</td>
+<td>
+
+```json
+{
+  "resource": { },
+  "attributes": { },
+  "body": {
+    "message": "some message",
+  },
+  "scope_name": "com.example.Foo",
+}
+```
+
+</td>
+</tr>
+</table>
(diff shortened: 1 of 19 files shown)


docs/types/logger_name.md
```diff
@@ -0,0 +1,64 @@
+## Logger Name Parsing
+
+A Logger Name may be parsed from a log entry in order to indicate the code from which a log was emitted. 
+
+### `logger_name` parsing parameters
+
+Parser operators can parse a severity and attach the resulting value to a log entry.
```
tigrannajaryan, line 7: I am confused, why does this refer to `severity?

entry/entry.go
```diff
@@ -51,6 +51,7 @@ type Entry struct {
 	TraceId      []byte            `json:"trace_id,omitempty"      yaml:"trace_id,omitempty"`
 	TraceFlags   []byte            `json:"trace_flags,omitempty"   yaml:"trace_flags,omitempty"`
 	Severity     Severity          `json:"severity"                yaml:"severity"`
+	LoggerName   string            `json:"logger_name"             yaml:"logger_name"`
```
tigrannajaryan, line 54: Do we want this to be called LoggerName or ScopeName? It is called InstrumentationScope Name in the data model.

operator/helper/parser_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
+// you may not use this file expect in compliance with the License.
```
tigrannajaryan, line 4: Unintented change?

entry/entry.go
djaglowski: This raises a question for me. In the context of logs, what value(s), if any, would reasonably be used as the scope, other than a logger name? Perhaps the log file name or similar? We could consider each input operator separately. 

If in the context of Logs, logger name is really the only sensible scope, then I would prefer to leave it as LoggerName to encourage use of only that specific value.

I need to think on this more, but at least in the case of file name, I don't think this is really the right value, as it really is a medium and not a scope from which the log was emitted. Perhaps some of the more native inputs like journald, windows event log have a sensible default though.

operator/helper/parser_test.go
djaglowski: Yes, thanks for catching it

docs/types/scope_name.md
```diff
@@ -0,0 +1,64 @@
...
+### Example Configurations
+
+#### Parse a scope_name from a string
+
+Configuration:
+```yaml
+- type: regex_parser
+  regexp: '^(?P<scope_name_field>\S*)\s-\s(?P<message>.*)'
+  scope_name:
+    parse_from: scope_name_field
```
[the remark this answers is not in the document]
jsirianni: I suppose in this base it would be `body.scope_name_field`

operator/helper/testdata/scope_name/preserve_to.yaml
```diff
@@ -0,0 +1,2 @@
+parse_from: $.from
```
jsirianni, line 1:
```suggestion
parse_from: body.from
```

docs/types/scope_name.md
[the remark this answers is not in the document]
djaglowski: Good catches. I'll correct these on this PR and make a followup PR for any other instances.

jsirianni (approved):

PR #401, opened 2022-02-18
Change attributes to support arbitrary data

chaitanyaphalak:
Partially addresses #190

### DIFF
--- a/entry/attribute_field.go
+++ b/entry/attribute_field.go
@@ -36,7 +36,7 @@ func (l AttributeField) Get(entry *Entry) (interface{}, bool) {
 // Set will set the attribute value on an entry
 func (l AttributeField) Set(entry *Entry, val interface{}) error {
 	if entry.Attributes == nil {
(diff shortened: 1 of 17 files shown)


djaglowski:
Thanks @chaitanyaphalak. This change looks good to me. 

However, [(contrib) #8022](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/8022) was merged, so the contrib failure we see now is a direct result of these changes, though I think not a problem with your PR. Still, I need to work through the necessary contrib changes before approving.

`./converter.go:350:27: cannot use v (type interface {}) as type string in argument to attributes.InsertString: need type assertion`

djaglowski:
@chaitanyaphalak I would like to merge this at roughly the same time as the corresponding changes to Resource. Are you working on those changes?

chaitanyaphalak:
yes @djaglowski the PR is here - https://github.com/open-telemetry/opentelemetry-log-collection/pull/411

PR #413, opened 2022-02-23
fix iteration

### LINKED ISSUE #391: data duplication at the beginning of log file
during perf test, I often see that the begging of log is consumed twice. (first 300~500 events out of 200,000) 
I will be looking into it. If anyone has an idea on how to fix it, please share.

rockb1017:
Fixes #391 
with this change, data duplication is gone. 
it is iterating over an array, but we are removing elements from that very array, so it causes to skip the next element after removing.

### DIFF
--- a/operator/input/file/file.go
+++ b/operator/input/file/file.go
@@ -199,30 +199,28 @@ func (f *InputOperator) makeReaders(filesPaths []string) []*Reader {
 	// Exclude any empty fingerprints or duplicate fingerprints to avoid doubling up on copy-truncate files
 OUTER:
 	for i := 0; i < len(fps); i++ {
-		if i < len(fps) {
-			fp := fps[i]
-			if len(fp.FirstBytes) == 0 {
+		fp := fps[i]
+		if len(fp.FirstBytes) == 0 {
+			if err := files[i].Close(); err != nil {
+				f.Errorf("problem closing file", "file", files[i].Name())
+			}
+			// Empty file, don't read it until we can compare its fingerprint
+			fps = append(fps[:i], fps[i+1:]...)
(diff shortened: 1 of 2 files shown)


rockb1017:
image

operator/input/file/file.go
```diff
@@ -199,25 +199,30 @@ func (f *InputOperator) makeReaders(filesPaths []string) []*Reader {
...
-			}
-			// Empty file, don't read it until we can compare its fingerprint
-			fps = append(fps[:i], fps[i+1:]...)
-			files = append(files[:i], files[i+1:]...)
-		}
-		for j := i + 1; j < len(fps); j++ {
-			fp2 := fps[j]
-			if fp.StartsWith(fp2) || fp2.StartsWith(fp) {
-				// Exclude
+		if i < len(fps) {
```
djaglowski, line 202: This exact condition is evaluated in the line above: `for i := 0; i < len(fps); i++ {`, so I don't think the result could possibly be different.

djaglowski (commented):
@rockb1017 This is a great find and fix. 

Two things:
1. See comment on the `if` condition. If I'm right, we can remove it and have a much clearer diff.
2. Do you have any way to reproduce the unwanted behavior in tests? I would hate for a future change to undo this fix. (I acknowledge this function is doing too much and should be pulled apart to make it more testable, but perhaps a higher level test can catch it for now.)

rockb1017:
i have applied your suggestion. Thank you. 
And I have added a test case.
When i comment out `i--` line, the test fails, but with it, it doesn't fail.

djaglowski (approved):
LGTM. Thanks @rockb1017

PR #416, opened 2022-02-25
feat(operator/recombine): do not combine logs before first_entry matches line

### LINKED ISSUE #415: recombine: do not combine if is_first_entry never matched
In case `is_first_entry` do not matches any log line, it will recombine full file into one log.
I would expect that `is_first_entry` flushes every line separately until it matches first line

configuration:
```
receivers:
  filelog/containers:
    include:
    - *.log
    operators:
    - type: regex_parser
      regex: '^(?P<timestamp>[^\s]+) (?P<stream>\w+) (?P<logtag>\w) (?P<message>.*)'
    - combine_field: message
      is_first_entry: $$body.message matches "^\\d{4}-\\d{1,2}-\\d{1,2}.\\d{2}:\\d{2}:\\d{2}.*"
      type: recombine
    start_at: beginning
```

logs:
```
2016-10-06T00:17:09.669794202Z stdout F non-matching line
2016-10-06T00:17:09.669794202Z stdout F another non-matching line
2016-10-06T00:17:09.669794202Z stdout F 2016-10-06T00:17:09.669794202Z matching line
```

output (shortened):
```
otelcol_1  | LogRecord #0
otelcol_1  | Body: {
otelcol_1  |      -> message: STRING(non-matching line
otelcol_1  | another non-matching line)
otelcol_1  | }
otelcol_1  | 2022-02-25T11:18:37.837Z   DEBUG   loggingexporter/logging_exporter.go:79  ResourceLog #0
otelcol_1  | Body: {
otelcol_1  |      -> message: STRING(2016-10-06T00:17:09.669794202Z matching line)
otelcol_1  | }
```

I would rather expect that `non-matching line` and `another non-matching line` as they aren't continuation of any previous log

sumo-drosiek:
Fixes #415 

do not combine logs before first_entry matches line
See attached issue for more details

### DIFF
--- a/operator/transformer/recombine/recombine.go
+++ b/operator/transformer/recombine/recombine.go
@@ -120,7 +120,6 @@ func (c *RecombineOperatorConfig) Build(logger *zap.SugaredLogger) (operator.Ope
 		ticker:              time.NewTicker(c.ForceFlushTimeout),
 		chClose:             make(chan struct{}),
 		sourceIdentifier:    c.SourceIdentifier,
-		firstLineMatched:    make(map[string]bool),
 	}, nil
 }
 
@@ -139,9 +138,6 @@ type RecombineOperator struct {
 	forceFlushTimeout   time.Duration
 	chClose             chan struct{}
 	sourceIdentifier    entry.Field
-
-	// firstLineMatched keeps information if any entry matched first line
-	firstLineMatched map[string]bool
 
 	sync.Mutex
 	batchMap map[string][]*entry.Entry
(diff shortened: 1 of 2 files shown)


operator/transformer/recombine/recombine.go
```diff
@@ -220,17 +224,26 @@ func (r *RecombineOperator) Process(ctx context.Context, e *entry.Entry) error {
...
+			// Flush the existing batch
+			err := r.flushSource(s)
+			if err != nil {
+				return err
+			}
+
+			// Add the current log to the new batch
+			r.addToBatch(ctx, e, s)
+			return nil
+		}
```
djaglowski, line 246: Do we really need to keep track of `firstLineMatched`? 

Can we just make sure never to start a batch with a partial (when matching on first line)?

```go
switch {
case matches && r.matchIndicatesFirst():
	... 
	return nil
case matches && r.matchIndicatesLast():
	... 
	return nil
case !matches && r.matchIndicatesFirst():
	if len(batchMap[s]) == 0 {
		// When matching on first entry, never batch partial first. Just emit immediately
		r.Write(context.Background(), e)
		return nil
	}
}
r.addToBatch(ctx, e, s)
```

sumo-drosiek: Yees, it's much better
Implemented :)

operator/transformer/recombine/recombine.go
```diff
@@ -231,10 +232,16 @@ func (r *RecombineOperator) Process(ctx context.Context, e *entry.Entry) error {
...
-	if matches && r.matchIndicatesLast() {
+	case matches && r.matchIndicatesLast():
+		r.addToBatch(ctx, e, s)
+		err := r.flushSource(s)
+		if err != nil {
+			return err
+		}
+		return nil
+	// When matching on first entry, never batch partial first. Just emit immediately
+	case !matches && r.matchIndicatesFirst() && len(r.batchMap[s]) == 0:
```
djaglowski, line 244: This case wound up with the exact same logic as the one above. I think in this case though, we do not need to mess with batching and flushing, and instead can just do `r.Write(context.Background(), entry)`.

I don't feel strongly about this, but at least we should deduplicate this code. Could just be:
```suggestion
	case matches && r.matchIndicatesLast():
		fallthrough
	// When matching on first entry, never batch partial first. Just emit immediately
	case !matches && r.matchIndicatesFirst() && len(r.batchMap[s]) == 0:
		r.addToBatch(ctx, e, s)
		return r.flushSource(s)
```

djaglowski (commented):
Sorry for the delayed response - on PTO last week.

This looks a lot better but I think needs one more small iteration to tidy it up.

djaglowski (approved):
LGTM. Thanks @sumo-drosiek

PR #425, opened 2022-03-09
Add support for parsing multiline csv records

### LINKED ISSUE #312: CSV parser does not handle multiline entries
The CSV parser is unable to handle CSV entries that span multiple lines, when a field of the CSV contains newline characters.

djaglowski:
The csv parser previously was unable to properly handle multiline values.
This change adds support. Newlines within csv records are preserved.

Resolves #312

### DIFF
--- a/operator/parser/csv/csv_test.go
+++ b/operator/parser/csv/csv_test.go
@@ -65,14 +65,14 @@ func TestCSVParserByteFailure(t *testing.T) {
 	parser := newTestParser(t)
 	_, err := parser.parse([]byte("invalid"))
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "record on line 1: wrong number of fields")
+	require.Contains(t, err.Error(), "wrong number of fields")
 }
 
 func TestCSVParserStringFailure(t *testing.T) {
 	parser := newTestParser(t)
 	_, err := parser.parse("invalid")
 	require.Error(t, err)
-	require.Contains(t, err.Error(), "record on line 1: wrong number of fields")
+	require.Contains(t, err.Error(), "wrong number of fields")
 }
 
 func TestCSVParserInvalidType(t *testing.T) {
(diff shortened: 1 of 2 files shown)


operator/parser/csv/csv_test.go
```diff
@@ -572,29 +572,176 @@ func TestParserCSV(t *testing.T) {
...
+				"D": "dddd",
+				"E": "eee\ne",
+			},
+		},
+		{
+			"everything",
+			"\n\na\na\n\naa,\n\nbb\nbb\n\n,cc\ncc\n\n,\ndddd\n,eeee\n\n",
+			map[string]interface{}{
+				"A": "a\na\naa",
+				"B": "\nbb\nbb\n",
```
atoulme, line 718: shouldn't this value end with two `\n`?

atoulme, line 718: shouldn't this value end with two `\n`?

operator/parser/csv/csv_test.go
```diff
@@ -572,29 +572,176 @@ func TestParserCSV(t *testing.T) {
...
+				"C": "cc\ncc\n",
+				"D": "\ndddd\n",
+				"E": "eeee",
```
atoulme, line 721: same, should have two `\n` at the end of the value

atoulme, line 721: same, should have two `\n` at the end of the value

operator/parser/csv/csv_test.go
```diff
@@ -572,29 +572,176 @@ func TestParserCSV(t *testing.T) {
...
-		cfg.Header = testHeader
+func TestParserCSVMultiline(t *testing.T) {
+	cases := []struct {
+		name     string
+		input    string
+		expected map[string]interface{}
+	}{
+		{
+			"first_field",
+			"aa\naa,bbbb,cccc,dddd,eeee",
```
atoulme, line 583: Can we test with no newlines as well, and several lines? I would like to make sure existing parsing still works

atoulme, line 583: Can we test with no newlines as well, and several lines? I would like to make sure existing parsing still works

atoulme: Can we test with escaped double quotes? Say

```"foo","bar","""foobar
is the new foo-bar""","another entry"```

atoulme: Can we test with escaped double quotes? Say

```"foo","bar","""foobar
is the new foo-bar""","another entry"```

djaglowski: I've added some more test cases mixing quotes and returns.

djaglowski: I added a basic case to this set.

operator/parser/csv/csv_test.go
djaglowski: Actually, not in this case, but I added a case where the empty line is preserved. This is actually covered in [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180) `Fields containing line breaks (CRLF), double quotes, and commas should be enclosed in double-quotes`

atoulme: makes sense

atoulme (approved):

jsirianni (approved):

PR #429, opened 2022-03-10
Remove metadata operator

djaglowski:
This operator assumes that resource and attributes are flat maps.
The add operator has made it redundant anyways.


djaglowski:
The `contrib-tests` CI failure is expected because contrib attempts to import the operator.

PR #430, opened 2022-03-10
Update Changelog ahead of major set of breaking changes

djaglowski:
This PR is a proposal for a release that would contain several breaking changes. 

If this PR is approved, the following PRs will be merged, and a new release will be cut:
- #364
- #370
- #371
- #372
- #397
- #429


CHANGELOG.md
```diff
@@ -5,13 +5,41 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## [0.26.1] - 2022-03-10
+## [0.90.0] - 2022-03-14
```
djaglowski, line 8: I specified v0.90.0 as a placeholder. My intention is to indicate that this is a significant change. Perhaps this is excessive, but I believe it would be effective.

djaglowski, line 8: The specified version and date are placeholders. 

I think these changes may warrant a more than a minor version bump, to indicate the significance.

djaglowski:
I believe these changes represent all of the breaking changes that have been agreed upon. However, the overall effort to align this library with the logs data model will include some additive changes, such as enhancements to field syntax. 

Additionally, #431 would be a notable breaking change but has not yet been agreed upon. Implementation of that depends on the enhancements to field syntax, and so may take a couple weeks. I'm inclined to cut this release sooner and if necessary a second breaking release in a couple weeks.

cc: @tigrannajaryan @jsirianni @pmm-sumo @dmitryax

dmitryax:
Do we have a chance to make all the changes (or at least some of them) in a backward compatible way whenever it's feasible?

For example, can https://github.com/open-telemetry/opentelemetry-log-collection/pull/364 support both `$body` and `body` for now, throwing warning if it's configured with `$`?

dmitryax:
Also it seems like we are accumulating all of them for one release. What is the reason for that? As a user, it would rather prefer adopting smaller amount of changes between upgrades rather than all at once, but others may not agree with me.

djaglowski:
> Also it seems like we are accumulating all of them for one release. What is the reason for that? As a user, it would rather prefer adopting smaller amount of changes between upgrades rather than all at once, but others may not agree with me.

This strategy was agreed in the Log SIG several weeks ago. I see tradeoffs either way, but I think this is the right approach. The risk with several consecutive breaking changes is that users have to keep revisiting and redeploying, and I think many would conclude at some point that it's not worth keeping up. At least with one, they have one painful experience and hopefully don't have to keep thinking about it. Besides, I think at this point it would introduce quite some delays to spread them out over several releases (of the collector).

djaglowski:
> Do we have a chance to make all the changes (or at least some of them) in a backward compatible way whenever it's feasible?
> 
> For example, can #364 support both `$body` and `body` for now, throwing warning if it's configured with `$`?
> 
> Sorry for bringing it too late.

Yeah, I think that one seems reasonable. I'll confirm and update it unless I'm forgetting a reason why it can't.

I'm not sure any of the others really can be made backwards compatible. I suppose technically #372 can add config flags and deprecate the old ones, but I think it's not worth it in that case.

jsirianni:
I think it would be nice to write a quick doc somewhere showing how to update existing configs. I would be happy to help with this, I am not sure where this doc would live. Perhaps the changelog / release.

Example:
- $body --> body
- how to handle entry.Timestamp now that it does not have a default
- restructure --> add / move / etc
- etc

djaglowski:
@jsirianni, that's a good idea. I tried to be verbose in the changelog, but something more end-user oriented would be great. If you want to rough out how you think that info would be best presented, I'm happy to help fill in the details. I think the best place for this info would be on the release page.

djaglowski:
I've updated #364 to allow the former syntax as well. Unfortunately, there's not a great way to complain about the deprecated usage because it is processed during unmarshaling, where we don't yet have a handle to a logger.

djaglowski:
@jsirianni, here is my draft for a special section at the top of the release notes. PTAL

# Upgrading to version `0.90.0`

Several changes have been made that affect configuration for the `filelog`, `syslog`, `tcplog`, `udplog`, and `journald` receivers.

### Update all usages of [field](/docs/types/field.md) syntax
  
Field syntax no longer requires uses the `$` character. Instead each field must begin with `body`, `attributes`, or `resource`.
  
<table>
<tr><th>Deprecated Example</th><th>Updated Equivalent</th></tr>
<tr><td>

`$attributes.["log.file.name"]`

</td><td>

`attributes.["log.file.name"]`

</td></tr><tr><td>

`$resource.["host.name"]`

</td><td>

`resource.["host.name"]`

</td></tr><tr><td>

`$body.foo`

</td><td>

`body.foo`

</td></tr><tr><td>

`$.foo`

</td><td>

`body.foo`

</td></tr><tr><td>

`foo`

</td><td>

`body.foo`

</td></tr>
</table>

### Replace usages of `restructure` operator

The `restructure` operator has been removed. Use `add`, `copy`, `flatten`, `move`, `remove`, and `retain` operators instead.

<table>
<tr><th>Deprecated Example</th><th>Updated Equivalent</th></tr>
<tr><td valign="top">

```yaml
operators:
  - type: restructure
    ops:
      - add:
        field: set_me
        value: foo
      - add:
        field: overwrite_me
        value: bar
      - move:
        from: details.env
        to: env
      - remove:
        field: delete_me
```

</td><td valign="top">

```yaml
operators:
  - type: add
    field: body.set_me
    value: foo
  - type: add
    field: body.overwrite_me
    value: bar
  - type: move
    from: body.details.env
    to: body.env
  - type: remove
    field: body.delete_me
```

</td></tr>
</table>

### Replace usages of `metadata` operator

The `metadata` operator has been removed. Use `add`, `copy`, or `move` operators instead.

<table>
<tr><th>Deprecated Example</th><th>Updated Equivalent</th></tr>
<tr><td valign="top">

```yaml
operators:
  - type: metadata
    attributes:
      environment: production
      file: 'EXPR( $body.file )'
    resource:
      cluster: blue
```

</td><td valign="top">

```yaml
operators:
  - type: add
    field: attributes.environment
    value: production
  - type: copy
    from: body.file
    to: attributes.file
  - type: add
    field: resource.cluster
    value: blue
  - type: move
    from: body.foo
    to: attributes.bar
```

</td></tr>
</table>

### Update `filelog` attribute references

The `filelog` receiver has adopted slightly different attribute names in order to match newly established semantic conventions. Configurations that previously refered to the `file.*` attributes should be updated.

<table>
<tr><th>Deprecated Attribute</th><th>Updated Equivalent</th></tr>
<tr><td>

`file.name`

</td><td>

`log.file.name`

</td></tr><tr><td>

`file.path`

</td><td>

`log.file.path`

</td></tr><tr><td>

`file.name.resolved`

</td><td>

`log.file.name_resolved`

</td></tr><tr><td>

`file.path.resolved`

</td><td>

`log.file.path_resolved`

</td></tr>
</table>


## **Note to Vendors**: A log record's `Timestamp` field may now be `nil`

[A recent change to the Log Data Model](https://github.com/open-telemetry/opentelemetry-specification/pull/2184) has redefined the usage of the `Timestamp` field. Correspondingly, this field is no longer initialized by default. All Log Exporters should be evaluated to ensure this change is handled accordingly.

Log exporters can use the following logic to mimic the previous funcationality (psuedocode):

```go
timestamp := log.ObservedTimestamp
if log.Timestamp != nil {
  timestamp = log.Timestamp
}
```

jsirianni:
@djaglowski looks great, exactly what I had in mind.

djaglowski:
@open-telemetry/log-collection-approvers, PTAL. Any additional feedback on this proposal? Can we move forward?

I would like to merge these PRs starting Wednesday if there are no objections.

djaglowski:
Good catch @tigrannajaryan. Fixed.

tigrannajaryan:
LGTM.

@dmitryax you had concerns about this change, PTAL.

pmm-sumo (approved):

dmitryax (approved):
LGTM

djaglowski:
All prerequisite PRs are now merged. 

Before this release is cut, I would like to call attention to the one remaining proposal that would be a breaking change. If #431 is accepted, it will ultimately require some users to perform a second update to their configurations. In many cases this will apply to lines that have just been updated due to this release. If this churn is coming and can be avoided, I think it is worth a slight delay in this release.

```yaml
# Pre-v0.28.0 Config
parse_from: sev

# Post-v0.28.0 Config
parse_from: body.sev

# Config if #431 is accepted
parse_from: attributes.sev
```

---

At this point, being late in the week, I'd like to release on Monday so I can be more available to support upgrading users. 

In the meantime, I would appreciate any additional thoughts on #431. I will work ahead on the implementation in case consensus can be reached quickly. We can update/validate the release timeline on Monday.

cc: @tigrannajaryan, @dmitryax, @jsirianni, @pmm-sumo

djaglowski:
I will cut this release this afternoon.

PR #434, opened 2022-03-15
fix(helpers/multiline): fix force flushing with multiline

sumo-drosiek:
I found an issue with multiline and force flushing. If `forcePeriod` is low, then sometimes logs buffer (data) have been pushed without proper multiline detection (just force flushed).

In order to fix that issue, I changed the `forceFlusher` to be oriented on changes inside `splitFunc`. It uses two variables to track changes:
* `lastDataChange` which keeps timestamp of last change (flush, new data, basically change in data)
* `previousDataLength` which keep length of data with `-1` as special variable (which means that data have been flushed, what prevents to force flush in case data changed but it length does not)

The splitFuncs are rather complicated, so I'm mostly relaying on unit tests

---

In order to reproduce the issue which started all of the changes

main.py
```python
import time

LOG='Mar 14 22:00:14 ip-11-11-11-111 sadd[25288]: asdasdasdfadfdsgasdgb asd asd asd sad asd asd asd as s'

with open('test.log', 'w') as f:
    for i in range(10):
        for i in range(5):
            f.write(LOG)
            f.write('\n')
        f.flush()
        time.sleep(0.3)

```

config.yaml
```yaml
receivers:
  filelog/syslog:
    start_at: beginning
    include_file_path_resolved: true
    include_file_name: false
    force_flush_period: 50ms
    max_log_size: 32MiB  # increase max_log_size to scrape long lines
    include:
      - ./test.log
    multiline:
      line_start_pattern: \w{3}\s+\d{1,2} \d{2}:\d{2}:\d{2}
    operators:
      - type: regex_parser
        regex: (?P<timestamp>\w{3}\s+\d{1,2} \d{2}:\d{2}:\d{2})
        preserve_to: $$body.log
        timestamp:
          parse_from: $$body.timestamp
          layout_type: gotime
          layout: Jan _2 15:04:05
      - type: restructure
        ops:
          - move:
              from: $$body.log
              to: $$body
    attributes:
      _a: syslog
      _b: syslog
      c: syslog
    resource:
      _d: otelcol
      e: otelcol
      f: otelcol
      g: otelcol
      h: otelcol
exporters:
  logging:
    logLevel: debug
service:
  pipelines:
    logs:
      receivers:
        - filelog/syslog
      exporters:
        - logging

```

### DIFF
--- a/operator/helper/multiline.go
+++ b/operator/helper/multiline.go
@@ -52,9 +52,9 @@ type Flusher struct {
 	lastDataChange time.Time
 
 	// previousDataLength:
-	// if < 0 - data has been flushed and we are waiting for new data
-	// if = 0 - no new data
-	// if > 0 - there is data which has not been flushed yet
+	// if previousDataLength < 0 - data has been flushed and we are waiting for new data
+	// if previousDataLength = 0 - no new data
+	// if previousDataLength > 0 - there is data which has not been flushed yet and it doesn't changed since lastDataChange
 	previousDataLength int
 }
 
@@ -74,9 +74,15 @@ func (f *Flusher) UpdateDataChangeTime(length int) {
 		return
 	}
 
-	// update internal properties with new values
+	// update internal properties with new values if data length changed
+	// because it means that data is flowing and being processed
 	f.previousDataLength = length
 	f.lastDataChange = time.Now()
+}
+
+// EnableForceFlush sets data length to 1 and doesn't touch lastDataChange
+func (f *Flusher) EnableForceFlush() {
+	f.previousDataLength = 1
 }
 
 // ShouldFlush returns true if data should be forcefully flushed
@@ -145,7 +151,6 @@ func (c MultilineConfig) getSplitFunc(encodingVar encoding.Encoding, flushAtEOF 
 // tokens that start with a match to the regex pattern provided
 func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) bufio.SplitFunc {
 	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
-		data = trimLeftWhitespaces(data)
 		firstLoc := re.FindIndex(data)
 		if firstLoc == nil {
 			// Flush if no more data is expected
@@ -153,17 +158,20 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
 				token = trimWhitespaces(data)
 				advance = len(data)
 				if force != nil {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 				}
 				return
 			}
 			if force != nil {
 				if force.ShouldFlush() {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 					token = trimWhitespaces(data)
 					advance = len(data)
 					return
 				} else {
+					// Inform flusher that we didn't flushed
 					force.UpdateDataChangeTime(len(data))
 				}
 			}
@@ -176,20 +184,25 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
 			// the beginning of the file does not match the start pattern, so return a token up to the first match so we don't lose data
 			advance = firstMatchStart
 			token = trimWhitespaces(data[0:firstMatchStart])
-			if force != nil {
-				force.UpdateDataChangeTime(-1)
+			if len(token) > 0 {
+				if force != nil {
+					// Inform flusher that we just flushed
+					force.UpdateDataChangeTime(-1)
+				}
+				return
 			}
-			return
 		}
 
 		if firstMatchEnd == len(data) {
 			if force != nil {
 				if force.ShouldFlush() {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 					token = trimWhitespaces(data)
 					advance = len(data)
 					return
 				} else {
+					// Inform flusher that we didn't flushed
 					force.UpdateDataChangeTime(len(data))
 				}
 			}
@@ -202,6 +215,7 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
 			token = trimWhitespaces(data)
 			advance = len(data)
 			if force != nil {
+				// Inform flusher that we just flushed
 				force.UpdateDataChangeTime(-1)
 			}
 			return
@@ -212,11 +226,13 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
 		if secondLoc == nil {
 			if force != nil {
 				if force.ShouldFlush() {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 					token = trimWhitespaces(data)
 					advance = len(data)
 					return
 				} else {
+					// Inform flusher that we didn't flushed
 					force.UpdateDataChangeTime(len(data))
 				}
 			}
@@ -228,6 +244,7 @@ func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) b
 		token = trimWhitespaces(data[firstMatchStart:secondMatchStart]) // the token begins at the first match, and ends at the beginning of the second match
 		err = nil
 		if force != nil {
+			// Inform flusher that we just flushed
 			force.UpdateDataChangeTime(-1)
 		}
 		return
@@ -263,17 +280,20 @@ func NewLineEndSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) buf
 				token = trimWhitespaces(data)
 				advance = len(data)
 				if force != nil {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 				}
 				return
 			}
 			if force != nil {
 				if force.ShouldFlush() {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 					token = trimWhitespaces(data)
 					advance = len(data)
 					return
 				} else {
+					// Inform flusher that we didn't flushed
 					force.UpdateDataChangeTime(len(data))
 				}
 			}
@@ -285,11 +305,13 @@ func NewLineEndSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) buf
 		if loc[1] == len(data)-1 && !atEOF {
 			if force != nil {
 				if force.ShouldFlush() {
+					// Inform flusher that we just flushed
 					force.UpdateDataChangeTime(-1)
 					token = trimWhitespaces(data)
 					advance = len(data)
 					return
 				} else {
+					// Inform flusher that we didn't flushed
 					force.UpdateDataChangeTime(len(data))
 				}
 			}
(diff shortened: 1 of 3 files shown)


operator/helper/multiline.go
```diff
@@ -155,21 +156,30 @@ func (c MultilineConfig) getSplitFunc(encodingVar encoding.Encoding, flushAtEOF
 // tokens that start with a match to the regex pattern provided
 func NewLineStartSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) bufio.SplitFunc {
 	return func(data []byte, atEOF bool) (advance int, token []byte, err error) {
-		if force != nil && force.ShouldFlush() {
-			force.Flush()
-			token = trimWhitespaces(data)
-			advance = len(data)
-			return
-		}
```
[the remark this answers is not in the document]
sumo-drosiek: if `data` contains multiple logs, they are going to be flushed as one log.

This is the issue which I wanted to fix primary

operator/helper/multiline.go
```diff
@@ -44,55 +44,56 @@ func (c *FlusherConfig) Build() *Flusher {
...
 }
 
-// CheckAndFlush sets internal flag to true if data is going to be force flushed
-func (f *Flusher) CheckAndFlush() {
-	if f.forcePeriod > 0 && time.Since(f.lastFlush) > f.forcePeriod && f.lastFlush.Sub(f.lastForcedFlush) > 0 {
-		f.force = true
-	}
+// Flushed reset data length
+func (f *Flusher) Flushed() {
+	f.UpdateDataChangeTime(-1)
```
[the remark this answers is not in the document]
sumo-drosiek: I wanted to avoid using `f.UpdateDataChangeTime(-1)` in several places (because of magic `-1`)

operator/helper/multiline.go
```diff
@@ -44,55 +44,56 @@ func (c *FlusherConfig) Build() *Flusher {
...
-	// lastFlush < lastForcedFlush => we just forced flush, set lastFlush to lastForcedFlush
-	lastFlush       time.Time
-	lastForcedFlush time.Time
+	// lastDataChange tracks date of last data change (including new data and flushes)
+	lastDataChange time.Time
+
+	// previousDataLength:
+	// if previousDataLength < 0 - data has been flushed and we are waiting for new data
+	// if previousDataLength = 0 - no new data
+	// if previousDataLength > 0 - there is data which has not been flushed yet and it doesn't changed since lastDataChange
```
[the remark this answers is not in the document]
sumo-drosiek: > How does `previousDataLength` tell us that we should or should not flush? I can't figure out how it matters at all.

`previousDataLength` has two functions:
  - prevents to force flush if it's <= 0 (it means that the flush has been performed recently or there is no data to flush)
  - keeps state of previous data length, so we can track if there was any change in processing

>Is there a meaningful difference when it has the value -1 vs 0?
there is no a meaningful difference, so I unified it

operator/helper/multiline.go
```diff
@@ -44,55 +44,56 @@ func (c *FlusherConfig) Build() *Flusher {
...
 
-// Flushed update lastFlush with current timestamp
-func (f *Flusher) Flushed() {
-	if f.lastFlush.Sub(f.lastForcedFlush) < 0 {
-		f.lastFlush = f.lastForcedFlush
-	} else {
-		f.lastFlush = time.Now()
+func (f *Flusher) UpdateDataChangeTime(length int) {
+	// Skip if length is greater than 0 and didn't changed
+	if length > 0 && length == f.previousDataLength {
```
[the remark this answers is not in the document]
sumo-drosiek: I think so, if client generates same log over and over again, and we are reading it simultaneously

sumo-drosiek:
Windows tests are failing due to time resolution issue. I'm thinking about another approach than reading from static string (as it not fully behaves like reading file)

sumo-drosiek:
@djaglowski I made some iteration over fixing/improving tests and not sure about something.

Should `multiline_test.go` be strictly unit tests or kind of integration tests. Strictly unit tests can be counterintuitive as now sometimes we do not expect all logs to be tokenized. In unit tests, if we got EOF and we didn't flushed everything it's fine.

This is fine, because if using file log receiver, after getting EOF, we are going to recreate scanner in some time and data will be flushed (as force flusher keeps state).

I have fixed myself on creating tests to be use case oriented, but it require them to be more like integration tests and then more complicated.

sumo-drosiek:
Also I refactored code and simplified the logic

operator/helper/multiline.go
```diff
@@ -44,55 +44,56 @@ func (c *FlusherConfig) Build() *Flusher {
...
+	f.UpdateDataChangeTime(-1)
 }
 
-// ForceFlushed update struct fields after forced flush
-func (f *Flusher) Flush() {
-	f.force = false
-	f.lastForcedFlush = time.Now()
+// EnableForceFlush sets data length to 1 and doesn't touch lastDataChange
+func (f *Flusher) EnableForceFlush() {
+	f.previousDataLength = 1
```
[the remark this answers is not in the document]
sumo-drosiek: According to tests, it is not needed after all

operator/helper/multiline.go
```diff
@@ -44,55 +44,56 @@ func (c *FlusherConfig) Build() *Flusher {
...
+
+	// previousDataLength:
+	// if previousDataLength < 0 - data has been flushed and we are waiting for new data
+	// if previousDataLength = 0 - no new data
+	// if previousDataLength > 0 - there is data which has not been flushed yet and it doesn't changed since lastDataChange
+	previousDataLength int
 }
 
-// NewFlusher Creates new Flusher with lastFlush set to unix epoch
+// NewFlusher Creates new Flusher with lastDataChange set to unix epoch
```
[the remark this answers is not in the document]
sumo-drosiek: Removed :+1:

djaglowski:
> @djaglowski I made some iteration over fixing/improving tests and not sure about something.
> 
> Should `multiline_test.go` be strictly unit tests or kind of integration tests. Strictly unit tests can be counterintuitive (but simple) as now sometimes we do not expect all logs to be tokenized. In unit tests, if we got EOF and we didn't flushed everything it's fine.
> 
> This is fine, because if using file log receiver, after getting EOF, we are going to recreate scanner in some time and data will be flushed (as force flusher keeps state).
> 
> I have fixed myself on creating tests to be use case oriented, but it require them to be more like integration tests and then more complicated. Overall I'm not sure if this is a good direction.

I'm in favor of writing tests that are more complex if that's what's necessary to validate the behavior.

operator/helper/multiline.go
```diff
@@ -226,16 +212,53 @@ func SplitNone(maxLogSize int) bufio.SplitFunc {
...
+		if token != nil {
+			if force != nil {
+				// Inform flusher that we just flushed
+				force.Flushed()
+			}
+			return
+		}
+
+		// If there is no token, force flush eventually
+		if force != nil {
```
djaglowski, line 239: To keep things flatter, we can just return here when necessary.

```suggestion
		if force == nil {
			return
		}	
```

operator/helper/multiline.go
```diff
@@ -226,16 +212,53 @@ func SplitNone(maxLogSize int) bufio.SplitFunc {
 	}
 }
 
-// NewLineEndSplitFunc creates a bufio.SplitFunc that splits an incoming stream into
-// tokens that end with a match to the regex pattern provided
-func NewLineEndSplitFunc(re *regexp.Regexp, flushAtEOF bool, force *Flusher) bufio.SplitFunc {
+func flusherSplitFunc(force *Flusher, splitFunc bufio.SplitFunc) bufio.SplitFunc {
```
djaglowski, line 215: Should be a receiver on the Flusher, instead of passing it in?

sumo-drosiek: I moved it and refactored `trimWhitespaces` and `getSplitFunc` to simplify usage of new function

sumo-drosiek:
> I'm in favor of writing tests that are more complex if that's what's necessary to validate the behavior.

Bugs which I found are covered now, and I hope that they are rather understandable with comments

I consider work on this PR as finished, but have in mind some next improvements:

* refactor `flushAtEOF` to be made in wrapper manner as force flusher now
* integration tests for filelog input

@djaglowski WDYT?

djaglowski:
@sumo-drosiek, thanks for your work on this. I apologize for the slow review. I'm taking this very cautiously because I have yet to master this area of the codebase, and because I haven't established confidence that we previously had sufficient test coverage to ensure no regressions. Taking another look now.

operator/helper/multiline_test.go
```diff
@@ -164,8 +160,39 @@ func TestLineStartSplitFunc(t *testing.T) {
 				"LOGPART log1\nLOGPART log1",
 			},
 			Flusher: &Flusher{
-				force:           true,
-				lastForcedFlush: time.Now(),
+				// We assume than in previous iteration we had same data length
```
djaglowski, line 163: I can't figure out why this is necessary.

operator/helper/multiline_test.go
```diff
@@ -164,8 +160,39 @@ func TestLineStartSplitFunc(t *testing.T) {
...
+				forcePeriod:        time.Second,
+			},
+		},
+		{
+			Name:    "LogsWithFlusherWithMultipleLogsInBuffer",
+			Pattern: `^LOGSTART \d+`,
+			Raw:     []byte("LOGPART log1\nLOGSTART 123\nLOGPART log1\t   \n"),
+			ExpectedTokenized: []string{
+				// We expect all logs except last one, as it will be flushed in next iteration
+				"LOGPART log1",
```
djaglowski, line 175: Shouldn't there be another token here?

sumo-drosiek: This is the perfect example for what I write before about strict unit tests being not so real life example.

This should be considered in file input operator point of view. File input operator reads data from file and run it against `splitFunc`. If there is some tokens to emit, it emits, otherwise it ends with some data unflushed. Filelog input operator is able to keep track pointer to last data being flushed. So in next iteration it takes data from that point (meantime can be more data added to the end of the file) and run against `splitFunc`. If there is no new data added to the end of the file for `forcePeriod`, flusher force to flush it.

This is hardly for me to explain, so I will add list just in case:

Filelog input operator:

1. Read file from previous pointer position (`0` for `start_at: beginning`, EOF for `start_at: end`, `pointer position` if already in loop
2. Run against `splitFunc` and update pointer position
  a. return tokens if any (will update pointer position to the end of last token)
  b. force flush if needed (no change since `forcePeriod`) (will update pointer position at the EOF)
  c. wait for more data (will do not update pointer position)
3. go to 1

---

So this test do not check `2b`, because in this unit case we can iterate only one, so there is no way to test `2a` and `2b` in current test and code design.

Code design limitation is that `2a` is considered as change and `lastDataChange` is updated to current timestamp, so `2b` will be skipped

operator/helper/multiline_test.go
sumo-drosiek: According to the explanation with list. This is `2b` case, so we have nth iteration of loop without any flush and new data added.

In other words, this unit case covers force flushing

sumo-drosiek:
> @sumo-drosiek, thanks for your work on this. I apologize for the slow review. I'm taking this very cautiously because I have yet to master this area of the codebase, and because I haven't established confidence that we previously had sufficient test coverage to ensure no regressions. Taking another look now.

@djaglowski This is completely fine. It took me time to get some understanding too. Also I performed some additional runs with custom builds to have more confidence that it behaves how I expect it

sumo-drosiek:
> 1. Create a custom implementation of `io.Reader` that behaves predictably in our test cases. Probably it takes a slice of timestamps and corresponding `[]byte` that are written at those times.

That is similar to what I was thinking of.

I created an issue so we can track further changes there: #448

PR #439, opened 2022-03-17
Enforce maximum SD-NAME length of 32, per RFC5424

djaglowski:
Longer SD-NAMEs were previously allowed in fork of go-syslog
dependency, but it is not clear that this is necessary.


djaglowski:
@pmm-sumo, @dmitryax, any opinions on this?

My thinking is that unless we have a clear and urgent need to make this specific exception to syslog format, we should use the upstream library.

pmm-sumo:
THe robustness principle says it's good to "be conservative in what you do and liberal in what you accept from" and I find it's always a good rule to follow in IT :) 

I think it would be better to support SD-NAME's that are beyond the RFC but then don't think following the standard strictly is going to do much harm. @sumo-drosiek any more thoughts on that?

pmm-sumo (approved):

dmitryax:
I don't know how popular this format is. If we remove its support, will we just drop the messages or they will be trimmed?

sumo-drosiek:
> I don't know how popular this format is. If we remove its support, will we just drop the messages or they will be trimmed?

I would leave test for `RFC5424LongSDName`, just update it according to the change

dmitryax (approved):
Looks good to me assuming that `RFC5424LongSDName` is widely used.

jsirianni (approved):

Issue #410, opened 2022-02-22
Field syntax behavior when "attributes" or "resource" are specified


djaglowski:
`entry.Entry.Attributes` and `entry.Entry.Resource` are being migrated from `map[string]string` to `map[string]interface{}` (See #190). A consequence of this change, is that "field syntax" must be enhanced such that users are able to refer to all fields contained within arbitrarily deep Attributes and Resource objects. 

For the most part, interacting with Attributes and Resource is the same as interacting with Body. Like the Body, there are special considerations when interacting with the "root" field. However, unlike the Body, the root field is a restricted data type (`map[string]interface{}`). This raises questions that are not addressed by the Body's implementation of the `entry.Field` interface. For example, is the following a valid configuration? If so, what is the expected behavior?

```yaml
type: copy
from: somewhere
to: attributes
```

This issue contains an analysis of how field syntax should behave for Attributes and Resource when interacting with the "root" field. It proposes some broad guidelines as a starting point, and then layers on specific implemenation recommendations for each operator or class of operators. The goal is to ensure robust and intuitive functionality.

## General Guidelines

1. Attributes and Resource should have the same or equivalent implementation of the `entry.Field` interface (`Get`, `Set`, `Delete`).
3. Operators should not rely on custom implementations of the Field interface. Rather, they should use _compositions_ of `Get`, `Set`, and `Delete`.
4. Operators may opt to disallow references to the root fields, for any or all field parameters.
5. All operators should clearly document, for each field parameter, the expected behavior when a root field is specified.
6. Field parameters should be validated at build time. Validity should not be dynamic. i.e. should not depend on the contents of a particular log entry. 

## `entry.Field` implementation for the "root" Attributes and Resource

`Get` should return the entire contents as `map[string]interface{}`.

`Set` should accept only `map[string]interface{}` and `map[string]string` as valid types. These should be "merged onto" the existing map. That is, this operation should be non destructive, except where there is a collision, in which case the old value is overwritten. (This is the current behavior for map values in the Body.)

`Delete` should replace the root field with `map[string]interface{}` and return the previous map.

## Specific Behaviors

The following addresses all current field parameters, across all operators in this library.

#### `write_to` 

All input operators allow users to specify where to write log contents. I have proposed elsewhere that this field be removed (See #408).

#### `syslog_input`

This operator contains a `syslog_parser`, which uses field parameters. The implementation of that parser will be inherited by this operator.

#### `parse_from` and `preserve_to`

[...]

PR #454, opened 2022-03-24
Allow field syntax to refer to arbitrary attribute and resource fields

### LINKED ISSUE #190: Resouce and Attributes should support arbitrary data

djaglowski:
Resolves #190

A couple notes on this implementation:
1. There's definitely room to refactor/extract logic. This is a verbose implementation but should be fully functional and performant.
2. This explicitly does not address #410, which will allow `resource` and `attributes` to be valid fields in their own right, in some cases. (The `retain` operator still has its preexisting special case for these.)
3. Many test cases were updated or added. In some cases, previous level of coverage was not great, so many of these are net new.


jsirianni (approved):
Working well for me. Example:
```yaml
receivers:
  filelog:
    include:
    - in
    start_at: beginning
    operators:
    - type: json_parser
      parse_from: body
      parse_to: resource.test
```

tigrannajaryan (approved):

Issue #431, opened 2022-03-10
Ideal parser settings


djaglowski:
Parsing behavior in this library was originally designed for a different data model than the one we now have. The expected usage of `body` and `attributes` have been redefined. As such, the behavior of parsers should be reevaluated.

This is not my first attempt to propose changes to the way parsers behave. I believe my previous attempts have been too focused on proposing _changes_, which is difficult to do without lengthy explanations of the current behavior. The following is instead meant to be read as a standalone proposal for how parsers should behave. Please read accordingly.

I suggest the following would be ideal behavior:
1. Parsing should be non-destructive by default.
2. Parsing should be destructive only when the same value is specified for both `parse_from` and `parse_to`.
3. The `parse_from` setting should default to `body`.
4. The `parse_to` setting should default to `attributes`.
5. The `preserve_to` setting should not exist.

## Rationale

1. Parsing should be non-destructive by default.

The primary purpose of parsing operations is to isolate information. Where the extracted information should be placed is a secondary, but obvious, concern. However, what happens to the original value is easily overlooked. As a general principle, destruction of the original value should require intentional configuration.

2. Parsing should be destructive only when the same value is specified for both `parse_from` and `parse_to`.

A somewhat common use case involves _applying_ structure to an unstructured log. For example, when a user extracts information from a line of text, they may wish keep only the key value pairs they have isolated. When this is the case, an intuitive choice is to overwrite the original value. This should be supported but should require explicit configuration.

3. The `parse_from` setting should default to `body`.

The body typically contains an unstructured log. Parsing this value is the most common parsing use case. Therefore, parsers should default to parsing the body.

4. The `parse_to` setting should default to `attributes`.

The result of a parsing operation is a structured object. This is true of all parsers that have a `parse_to` field.

The data model strongly encourages that structured data belong in `attributes`, so this is a natural default value for `parse_to`.

Additionally, the default value for `parse_to` should be different than that of `parse_from`, else we have a situation where parsing operations are destructive by default.

5. The `preserve_to` setting should not exist.

If parsing is non-destructive by default, there is no need to provide a special setting for preservation. In the case where a user wishes to "back up" a value and destroy the original, they can do so by copying or moving the value before parsing.

pmm-sumo:
@djaglowski FYI, I run the survey across several filelog receiver users and they support the changes

djaglowski:
Thank you @pmm-sumo, I'm glad to have many eyes on this!

[...]

PR #463, opened 2022-04-04
Change default value of 'parse_to' to 'attributes'

djaglowski:
Addresses one of two changes proposed in #431

### DIFF
--- a/operator/helper/parser.go
+++ b/operator/helper/parser.go
@@ -28,7 +28,7 @@ func NewParserConfig(operatorID, operatorType string) ParserConfig {
 	return ParserConfig{
 		TransformerConfig: NewTransformerConfig(operatorID, operatorType),
 		ParseFrom:         entry.NewBodyField(),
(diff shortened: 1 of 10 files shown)


djaglowski:
Contrib test failure is due to a test that expects parsing results to be present in Body rather than Attributes.

jsirianni (approved):

djaglowski:
@dmitryax, @tigrannajaryan, do you want to review this? It is a breaking change and currently only @pmm-sumo has written support on the proposal (#431). I think I recall hearing verbal agreement in last week's SIG, but want to make sure this is expected.

tigrannajaryan:
@djaglowski I will take a look, sorry for the delay.

tigrannajaryan (approved):
I am fine with https://github.com/open-telemetry/opentelemetry-log-collection/issues/431 so approving based on the description that this moves us in that direction. 
I don't know this codebase well to actually review the code, I trust the other reviewers who know the codebase much better.

PR #464, opened 2022-04-07
Remove preserve_to

### LINKED ISSUE #431: Ideal parser settings

djaglowski:
Resolves #431.

Removes the `preserve_to` setting from parsers.

Parsers no longer remove the `parse_from` field, unless promoting the value to a top-level field. (eg `timestamp`, `severity`, `scope`, `trace`)

### DIFF
--- a/operator/helper/parser.go
+++ b/operator/helper/parser.go
@@ -57,7 +57,6 @@ func (c ParserConfig) Build(logger *zap.SugaredLogger) (ParserOperator, error) {
 		TransformerOperator: transformerOperator,
 		ParseFrom:           c.ParseFrom,
 		ParseTo:             c.ParseTo,
-		PreserveTo:          c.PreserveTo,
 	}
 
 	if c.TimeParser != nil {
@@ -94,7 +93,6 @@ type ParserOperator struct {
 	TransformerOperator
 	ParseFrom       entry.Field
 	ParseTo         entry.Field
-	PreserveTo      *entry.Field
 	TimeParser      *TimeParser
 	SeverityParser  *SeverityParser
 	TraceParser     *TraceParser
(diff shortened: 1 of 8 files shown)


operator/helper/parser.go
```diff
@@ -82,6 +81,10 @@ func (c ParserConfig) Build(logger *zap.SugaredLogger) (ParserOperator, error) {
 		parserOperator.TraceParser = c.TraceParser
 	}
 
+	if c.ScopeNameParser != nil {
+		parserOperator.ScopeNameParser = c.ScopeNameParser
+	}
```
tigrannajaryan, line 86: Unrelated change?

djaglowski: Sort of. A bug I found while updating tests.

I'll split it into a separate PR.

djaglowski: https://github.com/open-telemetry/opentelemetry-log-collection/pull/465

jsirianni (approved):

PR #468, opened 2022-04-14
Patch v0.29.1 - Properly remove "preserve_to" setting

djaglowski:
This PR resolves two problems with the last release, both related to the usage of `preserve_to` in sub-parsers.
1. The setting was not fully removed from the codebase. It was removed from the base parser struct, but other usages of it existed in "sub-parsers". Additionally, many other parsers used it in their tests cases (via sub-parsers).
2. Likewise, the "preservation-by-default" behavior was not applied to sub-parsers. This behavior should be the same for all parsers in order to ensure a simple and clear rule across the board. It is also necessary, as it can be very difficult to preserve a value that is isolated by a json/regex/csv/kv parser, and then immediately consumed by a sub-parser.


djaglowski:
This patch was unplanned, but migrating the codebase over to contrib is going to take a little while yet. ([see comment here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9227#issuecomment-1099681639)) As a result, I think that cutting a patch release here makes sense. Hopefully this is the only one necessary.

jsirianni (approved):

Issue #476, opened 2022-04-26
Fix Windows Log Event XML parsing


armstrmi:
**Describe the bug**
Using the `windows input operator`, it has been noticed that with the introduction of Windows 2022, it has recently failed to receive some of the fields in a windows event log. Particularly `level`, `message`, `opcode`, `keywords`, etc. What appears to be the issue is that the input operator is attempting to parse out the XML format with a header of `RenderingInfo`. Unfortunately that `RenderingInfo` header only applies to events that have been collected by the [Windows Event Collector](https://docs.microsoft.com/en-us/windows/desktop/WEC/windows-event-collector), otherwise it simply does not exist in the `XML` of the log. For some unknown reason, it does not have this effect on previous versions of Windows server. Nonetheless, versions of this cannot be viewed. 

**Steps to reproduce**
1. Setup Input operator to read from `applications` channel in windows event log
2. Send an event log to the windows event log
3. attempt to read the newly sent logs using the windows Input operator

**What did you expect to see?**

I expected to see the full response of the windows event log, including all fields included in this xml format which is specified [here](https://github.com/open-telemetry/opentelemetry-log-collection/blob/main/operator/input/windows/xml.go#:~:text=type%20EventXML%20struct,%7D)
```xml
// EventXML is the rendered xml of an event.
type EventXML struct {
	EventID     EventID     `xml:"System>EventID"`
	Provider    Provider    `xml:"System>Provider"`
	Computer    string      `xml:"System>Computer"`
	Channel     string      `xml:"System>Channel"`
	RecordID    uint64      `xml:"System>EventRecordID"`
	TimeCreated TimeCreated `xml:"System>TimeCreated"`
	Message     string      `xml:"RenderingInfo>Message"`
	Level       string      `xml:"RenderingInfo>Level"`
	Task        string      `xml:"RenderingInfo>Task"`
	Opcode      string      `xml:"RenderingInfo>Opcode"`
	Keywords    []string    `xml:"RenderingInfo>Keywords>Keyword"`
}
```
```yaml
go test
1S C:\Users\xxx\Documents\gitfolder\opentelemetry-collector-contrib\receiver\windowseventlogreceiver>
map[channel:Application computer:xxxx event_id:map[id:10 qualifiers:0] keywords:[Classic] level:Error message:Test log opcode:Info provider:map[event_source: guid: name:EventCreate] record_id:26901 system_time:2022-04-22T19:07:04.687756100Z task:]
PASS
[...]

PR #478, opened 2022-05-03
Windows input fix

armstrmi:
**Changes**
- Implemented fix for the Windows Event Log Input Operator. 
- Main fix was to the `xml.go` file.
- Added tests in `xml_test.go` and a sample WEL xml file
 
**Link to Tracking Issue**
#476

### DIFF
--- a/operator/input/windows/xml.go
+++ b/operator/input/windows/xml.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build windows
 // +build windows
 
 package windows
@@ -32,11 +33,11 @@ type EventXML struct {
 	Channel     string      `xml:"System>Channel"`
 	RecordID    uint64      `xml:"System>EventRecordID"`
 	TimeCreated TimeCreated `xml:"System>TimeCreated"`
-	Message     string      `xml:"RenderingInfo>Message"`
-	Level       string      `xml:"RenderingInfo>Level"`
-	Task        string      `xml:"RenderingInfo>Task"`
-	Opcode      string      `xml:"RenderingInfo>Opcode"`
-	Keywords    []string    `xml:"RenderingInfo>Keywords>Keyword"`
+	Message     string      `xml:"EventData>Data"`
+	Level       string      `xml:"System>Level"`
+	Task        string      `xml:"System>Task"`
+	Opcode      string      `xml:"System>Opcode"`
+	Keywords    []string    `xml:"System>Keywords>Keyword"`
 }
 
 // parseTimestamp will parse the timestamp of the event.
--- a/operator/input/windows/xml.go
+++ b/operator/input/windows/xml.go
@@ -27,17 +27,22 @@ import (
 
 // EventXML is the rendered xml of an event.
 type EventXML struct {
-	EventID     EventID     `xml:"System>EventID"`
-	Provider    Provider    `xml:"System>Provider"`
-	Computer    string      `xml:"System>Computer"`
-	Channel     string      `xml:"System>Channel"`
-	RecordID    uint64      `xml:"System>EventRecordID"`
-	TimeCreated TimeCreated `xml:"System>TimeCreated"`
-	Message     string      `xml:"EventData>Data"`
-	Level       string      `xml:"System>Level"`
-	Task        string      `xml:"System>Task"`
-	Opcode      string      `xml:"System>Opcode"`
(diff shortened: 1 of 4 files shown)


operator/input/windows/xml.go
```diff
@@ -26,17 +27,22 @@ import (
...
+	TimeCreated      TimeCreated `xml:"System>TimeCreated"`
+	Message          string      `xml:"RenderingInfo>Message"`
+	RenderedLevel    string      `xml:"RenderingInfo>Level"`
+	Level            string      `xml:"System>Level"`
+	RenderedTask     string      `xml:"RenderingInfo>Task"`
+	Task             string      `xml:"System>Task"`
+	RenderedOpcode   string      `xml:"RenderingInfo>Opcode"`
+	Opcode           string      `xml:"System>Opcode"`
+	RenderedKeywords []string    `xml:"RenderingInfo>Keywords>Keyword"`
+	Keywords         []string    `xml:"System>Keywords"`
```
BinaryFissionGames, line 44: `Keywords` shouldn't be a slice, but a single string, right?

https://docs.microsoft.com/en-us/windows/win32/wes/eventschema-systempropertiestype-complextype

operator/input/windows/xml.go
```diff
@@ -26,17 +27,22 @@ import (
...
-	Keywords    []string    `xml:"RenderingInfo>Keywords>Keyword"`
+	EventID          EventID     `xml:"System>EventID"`
+	Provider         Provider    `xml:"System>Provider"`
+	Computer         string      `xml:"System>Computer"`
+	Channel          string      `xml:"System>Channel"`
+	RecordID         uint64      `xml:"System>EventRecordID"`
+	TimeCreated      TimeCreated `xml:"System>TimeCreated"`
+	Message          string      `xml:"RenderingInfo>Message"`
+	RenderedLevel    string      `xml:"RenderingInfo>Level"`
+	Level            string      `xml:"System>Level"`
```
djaglowski, line 38: Based on the issue description, it sounds like we might expect only one or the other of these. Is that right? If so, should we parse both here and only set whichever one is populated?

djaglowski (commented):
Can you add a test or two to demonstrate the functionality?

operator/input/windows/xml.go
armstrmi: That would be correct, only one of these would be populated, that being said I'm sure that could be done to only populate one of the fields. But I do have a question about that, would this mean getting rid of the `RenderedLevel` field of the EventXML struct, or simply removing that field from the output?

operator/input/windows/xml.go
armstrmi: I think it could be, I was going based off what was originally there, but, it does seem like `Keywords` could also be a uint64
```
        <xs:element name="Keywords"
            type="HexInt64Type"
            minOccurs="0"
         />
```

operator/input/windows/xml.go
djaglowski: I would expect that it only requires a change to the body of the log that is emitted.

BinaryFissionGames: Just to clarify, both may be populated. The `RenderingInfo` tag contains the "friendly" versions of these fields, for instance, `System>Level` may be `4`, but  `RenderingInfo>Level` would be `Informational`. In this case, both would be present on the XML returned from windows.

Only populating one field might make sense for the sake of removing redundancy, but I just want to clarify.

operator/input/windows/xml.go
```diff
@@ -26,17 +27,22 @@ import (
...
+	RenderedTask     string      `xml:"RenderingInfo>Task"`
+	Task             string      `xml:"System>Task"`
+	RenderedOpcode   string      `xml:"RenderingInfo>Opcode"`
+	Opcode           string      `xml:"System>Opcode"`
```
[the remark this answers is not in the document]
armstrmi: I think for now, if we are trying to remove redundancies, we should leave these as strings since it prevents any kind of type differences when populating the body. Since this can make the difference between setting a struct field as an int or a string.

operator/input/windows/xml.go
armstrmi: probably will just keep as string tho to reduce unneeded fields

armstrmi:
> Can you add a test or two to demonstrate the functionality?

@djaglowski there should be some tests in there now

BinaryFissionGames (approved):
This looks good to me,

djaglowski (approved):
Thanks @armstrmi

PR #486, opened 2022-05-16
tests(multiline): make tests more integration

Wolodija:
Fixes: #448

It's rather small refactor in order to achieve the aim than writing new tests


Wolodija:
I wasn't able to reproduce issue on macos. hope, the adjustment will work as expected

PR #489, opened 2022-05-18
`Fix`: Windows unsafe Pointers 

armstrmi:
## Description
The Windows input operator was having some issues with pointers when attempt to render an `EventXML`. Specifically the RenderSimple method. What was happening is that the `bufferUsed` pointer being passed into `evtRender` from `RenderSimple` was not being written to properly. This resulted in `Go` occasionally tossing aside the pointer before windows had a chance to write to it.

## Fix
 A fix we found is initializing `bufferUsed` in `evtRender` and then pass the pointer back to the calling function. This would still save the pointer as long as it was called to be used deeper in the stack.

## Testing
This newly changed repo was locally tested on a windows (2022) VM in which a `windowseventlog` receiver implements this piece of code for its log collection. The tests would write a Windows Event Log and have the receiver attempt to scrape those logs using the Windows input operator.

### DIFF
--- a/operator/input/windows/api.go
+++ b/operator/input/windows/api.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build windows
 // +build windows
 
 package windows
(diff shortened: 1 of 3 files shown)


djaglowski (approved):
I will merge this here because pkg/stanza is not yet hooked up, but will you also please open a PR with these changes against contrib? (pkg/stanza)

